PR #1, opened 2018-09-08
Models & Routing

govza:
Basic models for Surah & Ayah


fixtures/convert_to_fixtures.py
```diff
@@ -0,0 +1,20 @@
+
+'''convert surah.json to fixtures from
+https://github.com/semarketir/quranjson/tree/master/source'''
+
+import json
+
+with open('surah.json', encoding='utf-8') as data_file:
+    data = json.load(data_file)
+
+fixture_data = [
```
al-muammar: Why have you decided to call it `fixture`? It doesn't say me a lot about your data.

fixtures/surah_fixture.json
```diff
@@ -0,0 +1 @@
...h", "fields": {"surah_name": "Al-Qariah", "surah_ayah_count": 11}}, {"model": "surah.Surah", "fields": {"surah_name": "Al-Takathur", "surah_ayah_count": 8}}, {"model": "surah.Surah", "fields": {"surah_name": "Al-Asr", "surah_ayah_count": 3}}, {"model": "surah.Surah", "fields": {"surah_name": "Al-Humazah", "surah_ayah_count": 9}}, {"model": "surah.Surah", "fields": {"surah_name": "Al-Fil", "surah_ayah_count": 5}}, {"model": "surah.Surah", "fields": {"surah_name": "Quraish", "surah_ayah_count": 4}}, {"model": "surah.Surah", "fields": {"surah_name": "Al-Ma'un", "surah_ayah_count": 7}}, {"model": "surah.Surah", "fields": {"surah_name": "Al-Kauthar", "surah_ayah_count": 3}}, {"model": "surah.Surah", "fields": {"surah_name": "Al-Kafirun", "surah_ayah_count": 6}}, {"model": "surah.Surah", "fields": {"surah_name": "An-Nasr", "surah_ayah_count": 3}}, {"model": "surah.Surah", "fields": {"surah_name": "Al-Masad", "surah_ayah_count": 5}}, {"model": "surah.Surah", "fields": {"surah_name": "Al-Ikhlas", "surah_ayah_count": 4}}, {"model": "surah.Surah", "fields": {"surah_name": "Al-Falaq", "surah_ayah_count": 5}}, {"model": "surah.Surah", "fields": {"surah_name": "An-Nas", "surah_ayah_count": 6}}]
```
al-muammar: Why have you formatted JSONs everywhere except this place?

fixtures/convert_to_fixtures.py
```diff
@@ -0,0 +1,20 @@
+
```
al-muammar: Delete this line.

fixtures/convert_to_fixtures.py
```diff
@@ -0,0 +1,20 @@
...
+'''convert surah.json to fixtures from
```
al-muammar: I don't understand what are your fixtures. In Python in general scripts have the following outline:
```python
# Imports
import json

import click

# Functions implementing the main logic
def convert():
    pass

# Main function which parses command line arguments
@click.command()
@click.option('--count', default=1, help='number of greetings')
@click.argument('name')
def main(count, name):
    pass

if __name__ == "__main__":
    main()
```

pylintrc
```diff
@@ -0,0 +1,3 @@
+[BASIC]
```
al-muammar: Well, it's your local pylint. Committing this file if similar to committing some IDE files.
If you want to make things serious, explore https://travis-ci.org/
[StackOverflow question](https://stackoverflow.com/questions/12450348/how-to-integrate-any-python-lint-with-github-commit-status-api) about PyLint + GitHub.

surah/migrations/0002_auto_20180908_0913.py
```diff
@@ -0,0 +1,32 @@
+# Generated by Django 2.1 on 2018-09-08 09:13
```
al-muammar: Why do we need this migration? Can't you just update the first file?
Secondly, is it a good practice to update generated Django files? I'm not completely sure, but maybe.

surah/models.py
```diff
@@ -0,0 +1,10 @@
+# -*- coding: utf-8 -*-
```
al-muammar: Do you need it?

surah/models.py
```diff
@@ -0,0 +1,10 @@
...
+'''This module contains Surahs of Holy Quran'''
```
al-muammar: The description is not descriptive at all. It seems like we are storing some data in this file.

surah/models.py
```diff
@@ -0,0 +1,10 @@
...
+from django.db import models
+
+
+class Surah(models.Model):
+    '''Surah class with id name and number of ayahs'''
```
al-muammar: Too many details in this description.

surah/models.py
```diff
@@ -0,0 +1,10 @@
...
+    surah_id = models.AutoField(primary_key=True)
+    surah_name = models.CharField(max_length=250)
```
al-muammar: 1. Why 250? I don't remember so long names of surahs.
2. Write some help text.

surah/viewsets.py
```diff
@@ -0,0 +1,12 @@
+# -*- coding: utf-8 -*-
```
al-muammar: Is it a good practice to put views and viewsets into separate files? At least I'm not aware of it :)

al-muammar (changes requested):
LGTM, lots of new shiny stuff used which I'm not super aware of :)
1. Move fixtures to a separate PR
2. I don't like the usage of REST API at this point. At the current stage, I completely don't understand why do we need it. You can either write it in the description to the PR or make a separate one.

rguliev (changes requested):
In general I would agree with @Jihadik . But I would like to add a few points as well:

1) Abut fixtures. Sorry, I was not aware of that type of data initializing (for info: https://docs.djangoproject.com/en/2.1/howto/initial-data/). If you want to use this approach instead of migration - ok. But then make some fixes, pleace:
-   Your fixtures folder by default supposed to be inside the app folder. According to documentation: By default, Django looks in the fixtures directory inside each app for fixtures.
- Please format your json, because now it's difficult for view and for tracing changes, i.e. all changes (if they will be) will be at like 1. You do not have to do it by script actually. You can take existing file and put in some editor which can format it automatically.
- I would put all helpers and sources (i.e. surah.json and convert_to_fixtures.py in the current case) in separate folder(s) (for example, helpers and sources in the root dir). Do not mix helpers/sources with Django project's files/folders. I mean if fixtures by design is a folder for initial data, then use it just for initial data.

2) I also want to make it clear that MODEL != APP. Seems like you want to create an app for each model, like "surah" app for surah model/table, "ayah" app for ayah model/table, etc. So, I see no reason for creating an app just for one model. Usually the app is a subpath of URL, I mean 'admin' app for 'www.yourwebdomain.com/admin/\*', 'stats' for 'www.yourwebdomain.com/stats/\*'. It is a flexible rule you don't have to do so, but this one is the clearest way. So for now, you basically need to create one app (you can call it 'core' or 'main') and put there all models: surah, ayah, reciter, etc. If that was your actual plan, then I think that "surah" is not a good name for the app.

fixtures/convert_to_fixtures.py
govza: I thought it's common way to call files for initial data loading
https://code.djangoproject.com/wiki/Fixtures

fixtures/surah_fixture.json
govza: it's result of convert_to_fixtures.py file

fixtures/convert_to_fixtures.py
govza: Used this to convert external surah.json file from quranapiproject to fixtures files, which can populate our database. Kept this for futures additions, ec we decide to add arabic name of surah, juz, other data.

surah/migrations/0002_auto_20180908_0913.py
govza: Separating model creating and database seeding should be ok. Not auto created but with  makemigrations --empty yourappname

surah/models.py
govza: No :)

surah/models.py
govza: No much difference unless we go over 255 with another byte, don't know actually the longest name & wasn't sure about number of max chars.
descriptions is pain :(

fixtures/convert_to_fixtures.py
al-muammar: I’m not saying ”delete this script”, just write it in a proper way :)

fixtures/surah_fixture.json
al-muammar: It doesn't justify you! :)

surah/models.py
al-muammar: Each ascii symbol is 1 byte. Each utf-8 symbol consumes even more. I’m not completely sure how it's implemented in DBs, but it's a waste of resources.

0-255 is a range of values that you can encode with 1 byte. It's a different thing.

govza: Actually I found it's interesting that there no difference. In contrast to CHAR, VARCHAR values are stored as a 1-byte or 2-byte length prefix plus data. https://dev.mysql.com/doc/refman/8.0/en/char.html

govza: Ok but it impacts on temporary files, so I agree to change it according lenght

helpers/surah/convert_surah_json.py
```diff
@@ -0,0 +1,31 @@
...
+https://github.com/semarketir/quranjson/blob/master/source/surah.json
+to django fixtures file'''
+
+import json
+from os import path
+
+
+def main():
+    fixture_dir = path.abspath(
+        path.join(path.dirname(__file__), '../../quran/fixtures'))
```
al-muammar, line 11: Define this path in `setting.py` and don't use raw paths.

quran/admin.py
```diff
@@ -0,0 +1,3 @@
+from django.contrib import admin
```
al-muammar: Will admin work without this file? If yes, delete.

requirements.txt
```diff
@@ -1,4 +1,5 @@
 Django==2.1
 python-dotenv==0.7.1
 psycopg2==2.7.4
-gunicorn==19.7.1
\ No newline at end of file
+gunicorn==19.7.1
+djangorestframework==3.8.2
```
al-muammar: Remove this dependency.

requirements.txt
```diff
@@ -1,4 +1,5 @@
 Django==2.1
 python-dotenv==0.7.1
 psycopg2==2.7.4
-gunicorn==19.7.1
\ No newline at end of file
+gunicorn==19.7.1
```
al-muammar: Remove this dependency. This PR is about models, not gunicorns.

quranwords/wsgi.py
```diff
@@ -11,6 +11,6 @@
 
 from django.core.wsgi import get_wsgi_application
 
-os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'src.settings')
+os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'quranwords.settings')
```
al-muammar: Rename `quranwords` to `config`.

quranwords/routers.py
```diff
@@ -0,0 +1,7 @@
+# -*- coding: utf-8 -*-
```
al-muammar: Remove this file for now, we don't need these routers at the current stage.

quranwords/urls.py
```diff
@@ -14,8 +14,10 @@
     2. Add a URL to urlpatterns:  path('blog/', include('blog.urls'))
 """
 from django.contrib import admin
-from django.urls import path
+from django.urls import path, include
+from .routers import router
```
al-muammar: Remove

quranwords/urls.py
```diff
@@ -14,8 +14,10 @@
...
 
 urlpatterns = [
     path('admin/', admin.site.urls),
+    path('api/', include(router.urls))
```
al-muammar: Remove and instead add your app.

quranwords/settings.py
```diff
@@ -20,6 +20,8 @@
     'django.contrib.sessions',
     'django.contrib.messages',
     'django.contrib.staticfiles',
+    'quran',
+    'rest_framework'
```
al-muammar: Remove it for now.

quran/serializers.py
```diff
@@ -0,0 +1,13 @@
+# -*- coding: utf-8 -*-
```
al-muammar: Remove this file for now. Maybe we will need it, but later.

quran/models.py
```diff
@@ -0,0 +1,20 @@
+'''Models describing abstractions used at Quran app'''
+from django.db import models
+from django.contrib.postgres.fields import ArrayField
+
+
+class Surah(models.Model):
+    '''General information about Surahs'''
+    surah_id = models.AutoField(primary_key=True)
```
al-muammar: Name it just `id`.

quran/models.py
```diff
@@ -0,0 +1,20 @@
...
+    surah_name = models.CharField(max_length=20)
+    surah_ayah_count = models.PositiveIntegerField()
+
+
+class Ayah(models.Model):
+    '''General information about Ayah'''
+    ayah_id = models.AutoField(primary_key=True)
```
al-muammar: Name it just `id`.

quran/models.py
```diff
@@ -0,0 +1,20 @@
...
+    ayah_text = ArrayField(
```
al-muammar: Name it just `text`.

quran/models.py
```diff
@@ -0,0 +1,20 @@
...
+        models.CharField(max_length=255, blank=True, default=''),
```
al-muammar: Ayah's text can be quite large, we need to increase the size of this field. Or we should consider using `TextField` for it: https://stackoverflow.com/questions/7354588/django-charfield-vs-textfield

quran/models.py
```diff
@@ -0,0 +1,20 @@
...
+        default=list)
+    surah_id = models.ForeignKey(
+        'quran.Surah', on_delete=models.CASCADE, related_name='ayahs')
```
al-muammar: Change `'quran.Surah'` by `Surah`. Never ever use raw strings when you can avoid that.
`'ayahs'` -- ArabEnglish

quran/tests.py
```diff
@@ -0,0 +1,3 @@
+from django.test import TestCase
```
al-muammar: No tests, no code, remove this file.

quran/viewsets.py
```diff
@@ -0,0 +1,12 @@
+# -*- coding: utf-8 -*-
+'''Here combine the logic for a set of related views in a single class'''
+
+from rest_framework import viewsets
+from .models import Surah
+from .serializers import SurahSerializer
+
+
+class SurahViewSet(viewsets.ModelViewSet):
```
al-muammar: Move it to views. But now just remove it, your PR is about models, not views.

quran/migrations/0003_ayah.py
```diff
@@ -0,0 +1,34 @@
+# Generated by Django 2.1 on 2018-09-09 17:26
```
al-muammar: Move it to the first migration.

helpers/surah/convert_surah_json.py
```diff
@@ -0,0 +1,31 @@
...
+    fixture_filename = 'surah_data_init.json'
```
al-muammar: Define some constant for it.

helpers/surah/convert_surah_json.py
```diff
@@ -0,0 +1,31 @@
...
+    fixture_file = path.join(fixture_dir, fixture_filename)
+
+    with open('surah.json', encoding='utf-8') as data_file:
```
al-muammar, line 15: Define some constant for `'surah.json'`.

helpers/surah/convert_surah_json.py
```diff
@@ -0,0 +1,31 @@
...
+        data = json.load(data_file)
+
+    surah_data = [
```
al-muammar, line 18: Improve the script, so it populates not only your metadata but adds the real text of the ayaat.

quran/models.py
```diff
@@ -0,0 +1,20 @@
+'''Models describing abstractions used at Quran app'''
+from django.db import models
+from django.contrib.postgres.fields import ArrayField
+
+
+class Surah(models.Model):
+    '''General information about Surahs'''
+    surah_id = models.AutoField(primary_key=True)
+    surah_name = models.CharField(max_length=20)
```
al-muammar: Just `name`.

quran/models.py
```diff
@@ -0,0 +1,20 @@
...
+    surah_ayah_count = models.PositiveIntegerField()
```
al-muammar: Just `ayah_count` or `total_ayahs`.

al-muammar (changes requested):
There are lots of comments, but in general, just remove everything unrelated to this PR.

helpers/surah/convert_surah_json.py
govza: I think this is it https://github.com/it-muslim/QuranWords/pull/1/commits/d094bb02f71875b89e91ed10f88c761ff0da56d9
will mark as resoved

config/urls.py
```diff
@@ -15,7 +15,9 @@
 """
 from django.contrib import admin
 from django.urls import path
+from quran import views
 
 urlpatterns = [
     path('admin/', admin.site.urls),
+    path('', views.home, name='home')
```
al-muammar: This file should know only in general about your app, but url specifics should be in `quran.urls` file, an you need to do here something like:
```python
path('', include('quran.urls')),
```
For reference: https://docs.djangoproject.com/en/2.1/intro/tutorial01/#write-your-first-view

helpers/ayah/convert_quran_words_ar_csv.py
```diff
@@ -0,0 +1,68 @@
+'''convert quran_words_ar.csv to django fixtures file'''
```
al-muammar: Capitalize the first letter.

helpers/ayah/convert_quran_words_ar_csv.py
```diff
@@ -0,0 +1,68 @@
...
+import json
+from os import path
+import csv
+from pprint import pprint
+from collections import defaultdict
+
+
+def structure_surah_dict(data):
+    '''parse raw csv data and build structured dict'''
```
al-muammar: Ditto

helpers/ayah/convert_quran_words_ar_csv.py
```diff
@@ -0,0 +1,68 @@
...
+            surah_dict[surah][ayah] = []
+
+        if word not in surah_dict[surah][ayah]:
+            surah_dict[surah][ayah].append(word)
+
+    return surah_dict
+
+
+def structure_ayah_json(surah_dict):
+    '''form import ready ayah list json from structured dict'''
```
al-muammar: Ditto

helpers/ayah/convert_quran_words_ar_csv.py
```diff
@@ -0,0 +1,68 @@
...
+    fixture_file = path.join(fixture_dir, fixture_filename)
+
+    data = list()
+
+    with open('quran-words-ar.csv', encoding='utf-8') as data_file:
+        fieldnames = ("surah", "ayah", "position", "text")
+        reader = csv.DictReader(data_file, fieldnames)
+        for row in reader:
+            data.append([
+                row.get("surah"),
```
al-muammar: `int(row.get("surah")),`

helpers/ayah/convert_quran_words_ar_csv.py
```diff
@@ -0,0 +1,68 @@
...
+                row.get("ayah"),
```
al-muammar: ditto

helpers/ayah/convert_quran_words_ar_csv.py
```diff
@@ -0,0 +1,68 @@
...
+                row.get("position"),
```
al-muammar: ditto

helpers/ayah/convert_quran_words_ar_csv.py
```diff
@@ -0,0 +1,68 @@
...
+from pprint import pprint
+from collections import defaultdict
+
+
+def structure_surah_dict(data):
+    '''parse raw csv data and build structured dict'''
+    surah_dict = {}
+    for surah, ayah, position, word in data[1:]:
+
+        if surah not in surah_dict:
```
al-muammar: Use `defaultdict` which is defined in `collections`.

helpers/ayah/convert_quran_words_ar_csv.py
```diff
@@ -0,0 +1,68 @@
...
+            surah_dict[surah] = {}
+
+        if ayah not in surah_dict[surah]:
```
al-muammar: Use `defaultdict` here as well.

helpers/ayah/convert_quran_words_ar_csv.py
```diff
@@ -0,0 +1,68 @@
...
+            surah_dict[surah][ayah] = []
+
+        if word not in surah_dict[surah][ayah]:
```
al-muammar: This looks completely wrong to me. It means that no word should repeat in one ayah twice.
This code will corrupt some ayahs. Which is SUPER bad.

quran/models.py
```diff
@@ -0,0 +1,23 @@
...
+class Surah(models.Model):
+    '''General information about Surahs'''
+    id = models.AutoField(primary_key=True)
+    name = models.CharField(max_length=20)
+    total_ayahs = models.PositiveIntegerField()
+
+
+class Ayah(models.Model):
+    '''General information about Ayah'''
+    id = models.AutoField(primary_key=True)
```
al-muammar: This is the biggest problem. You can't declare an `id` to be a `primary_key`. `id`-s are **not unique**! That's why the highest number of ayah was 286. They overrode each other!
cc @rguliev

quran/models.py
```diff
@@ -0,0 +1,23 @@
...
+    text = ArrayField(
```
al-muammar: I found a quite nice solution to get rid of the dependency on a specific DB:
```python
    text = models.TextField(),

    def set_text(self, words):
        self.text = " ".join(words)

    def get_text(self, text):
        return text.split()
```
But don't forget to correct your script in this case.

quran/models.py
```diff
@@ -0,0 +1,23 @@
+'''Models describing abstractions used at Quran app'''
+from django.db import models
+import django.contrib.postgres.fields
+from django.contrib.postgres.fields import ArrayField
+
+
+class Surah(models.Model):
+    '''General information about Surahs'''
+    id = models.AutoField(primary_key=True)
```
al-muammar: Let's make everything uniform and remove `primary_key=True`.

al-muammar (changes requested):
The code is way better now, but there is still a bunch of work:
1. Let's go back to py-sqlite DB since it is nice for all kinds of experiments.
2. Correct your models, make them simpler.
3. Merge all migrations into one. That's fine! It is way easier to read.

config/urls.py
rguliev: And now we can see that `quran` is not a good name for the app containing a home page and all main routes...

quran/models.py
rguliev: I would stick with this one too, rather than use DB specific solution.

config/urls.py
govza: Its just dummy page I added, to not see 404 at index of project. I didn't quite into it right now, but that should be replaced with rest-framework views. 
http://www.django-rest-framework.org/api-guide/views/

rguliev: It's ok that you made it as a dummy for now. But we will need a homepage anyway. Honestly, I don't get it, what is wrong with simple Django views? They are more than enough for our goals.

quran/models.py
govza: Wasn't sure if words here real words divided by spaces, or some logical entities translated.

govza: Ok , just checked, there are also "word" with space as well 37,130,3,إِلْ يَاسِينَ  it's only one, but exists.
May be we can split comma separated words stored as TextField as well?

quran/models.py
govza: It's good to have primary key in every table, here surah ids are unique and referenced from ayah as ForeignKey, so my guess leave it.

al-muammar: Primary key is a MUST in every table and if you omit it’s specification, it will be created automatically in your model and called `pk`.

rguliev (approved):

helpers/ayah/convert_quran_words_ar_csv.py
```diff
@@ -0,0 +1,62 @@
...
+from os import path
+import csv
+from collections import defaultdict
+
+
+def structure_surah_dict(data):
+    '''Parse raw csv data and build structured dict'''
+    surah_dict = defaultdict(lambda: defaultdict(list))
+    for surah, ayah, position, word in data:
+        if ayah not in surah_dict[surah]:
```
al-muammar: You can delete this as well.

helpers/ayah/convert_quran_words_ar_csv.py
```diff
@@ -0,0 +1,62 @@
...
+            surah_dict[surah][ayah] = []
+
+        if position > len(surah_dict[surah][ayah]):
```
al-muammar: This condition doesn't make sense to me, tbh. The bad case if you will see the following two lines:
```
1, 1, 2, word_2
1, 1, 1, word_1
```
In this case you will throw away `word_1` which is incorrect.
Either remove this check completely and assume that the data is sorted by position, or make a proper fix.

helpers/ayah/convert_quran_words_ar_csv.py
```diff
@@ -0,0 +1,62 @@
...
+            surah_dict[surah][ayah].append(word)
+    return surah_dict
+
+
+def structure_ayah_json(surah_dict):
+    '''Form import ready ayah list json from structured dict'''
+    list_of_ayahs = []
+    for surah, surah_values in surah_dict.items():
```
al-muammar: Give your variables better names:
```python
for surah, ayahs in surah_dict.items():
    for ayah, words in ayahs.items():
        ...
```

helpers/ayah/convert_quran_words_ar_csv.py
```diff
@@ -0,0 +1,62 @@
...
+        for ayah, ayah_values in surah_values.items():
+            element = {
+                "model": "quran.Ayah",
+                "fields": {
+                    "surah": surah,
+                    "ayah": ayah,
+                    "text": ayah_values
+                }
+            }
+            list_of_ayahs.append(element.copy())
```
al-muammar: There is no need in copy, I believe.

helpers/ayah/convert_quran_words_ar_csv.py
```diff
@@ -0,0 +1,62 @@
...
+    '''Form import ready ayah list json from structured dict'''
+    list_of_ayahs = []
+    for surah, surah_values in surah_dict.items():
+        for ayah, ayah_values in surah_values.items():
+            element = {
+                "model": "quran.Ayah",
+                "fields": {
+                    "surah": surah,
+                    "ayah": ayah,
+                    "text": ayah_values
```
al-muammar: I'm not sure, but don't you need to join your words with a comma?

helpers/ayah/convert_quran_words_ar_csv.py
```diff
@@ -0,0 +1,62 @@
...
+    return list_of_ayahs
+
+
+def main():
+    fixture_dir = path.abspath(
+        path.join(path.dirname(__file__), '../../quran/fixtures'))
+    fixture_filename = 'ayah_list.json'
+    fixture_file = path.join(fixture_dir, fixture_filename)
+
+    data = list()
```
al-muammar:
```python
data = []
```

helpers/ayah/convert_quran_words_ar_csv.py
```diff
@@ -0,0 +1,62 @@
...
+
+    with open('quran-words-ar.csv', encoding='utf-8') as data_file:
+        fieldnames = ("surah", "ayah", "position", "text")
```
al-muammar:
```python
field_names
```

helpers/surah/convert_surah_json.py
```diff
@@ -0,0 +1,31 @@
...
+        {
+            "model": "quran.Surah",
+            "fields": {
+                "name": element['title'],
+                "total_ayahs": element['count']}
+        }
+        for element in data]
```
al-muammar:
```python
for surah in data
```

quran/models.py
```diff
@@ -0,0 +1,24 @@
...
+    name = models.CharField(max_length=20)
+    total_ayahs = models.PositiveIntegerField(default=0)
+
+
+class Ayah(models.Model):
+    '''General information about Ayah'''
+    id = models.AutoField(primary_key=True, serialize=False)
```
al-muammar: Don't convert an `id` to a `primary_key`. Just remove that `primary_key=True` and `serialize=False`.

quran/models.py
```diff
@@ -0,0 +1,24 @@
...
+    ayah = models.PositiveIntegerField(default=0)
+    text = ArrayField(
```
al-muammar: It is still an array field. Can you convert it to the proposed format, so everyone can easily set up this project and play with sqlite?

quran/models.py
```diff
@@ -0,0 +1,24 @@
...
+        models.CharField(max_length=255, blank=True),
```
al-muammar: Remove `blank=True`. Is there an empty word in the Qur'an? I don't know such words.

quran/models.py
```diff
@@ -0,0 +1,24 @@
...
+        blank=True,
```
al-muammar: Remove `blank=True`. There is no ayah with 0 words in it.

quran/models.py
```diff
@@ -0,0 +1,24 @@
...
+        null=True,)
```
al-muammar: The same here, remove it. An ayah MUST have a text!

al-muammar (changes requested):
You are close to the destination, but still, you need to correct a few things here and there.

quran/models.py
rguliev: You can actually remove the whole line. `id` field will be added automatically: https://docs.djangoproject.com/en/2.1/topics/db/models/#automatic-primary-key-fields

helpers/ayah/convert_quran_words_ar_csv.py
govza: Now they stored in TextField like string representation of array ['ٱهْدِنَا', 'ٱلصِّرَٰطَ', 'ٱلْمُسْتَقِيمَ'] 
Luckily there no commas alike in arabic, but  I'm not sure about comma separated strings behavior while serializing.

config/settings.py
```diff
@@ -20,6 +20,7 @@
     'django.contrib.sessions',
     'django.contrib.messages',
     'django.contrib.staticfiles',
+    'quran'
```
al-muammar: Don't forget to add trailing comma.
You should consider using automatic formatter.

helpers/ayah/convert_quran_words_ar_csv.py
```diff
@@ -0,0 +1,58 @@
...
+import json
+from os import path
+import csv
+from collections import defaultdict
+
+
+def structure_surah_dict(data):
+    '''Parse raw csv data and build structured dict'''
+    surah_dict = defaultdict(lambda: defaultdict(list))
+    for surah, ayah, position, word in data:
```
al-muammar, line 11: You see, now the code is way more concise!

quran/models.py
```diff
@@ -0,0 +1,18 @@
+'''Models describing abstractions used at Quran app'''
+from django.db import migrations, models
+
+
+class Surah(models.Model):
+    '''General information about Surahs'''
+    name = models.CharField(max_length=20)
```
al-muammar, line 8: You still need an `id` field. In this case, defining it as a primary key was totally fine.
Otherwise, how are you going to query surah number `10`, let's say? You can't rely on the fact that your primary key would coincide with the real numbers of surahs.

quran/models.py
```diff
@@ -0,0 +1,18 @@
...
+    total_ayahs = models.PositiveIntegerField(default=0)
```
al-muammar: This default doesn't make much sense to me. Remove it.

quran/models.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class Ayah(models.Model):
+    '''General information about Ayah'''
+    ayah = models.PositiveIntegerField(default=0)
```
al-muammar: Add an `id` field. How else are you going to sort ayahs, when a frontend will make API call to get some specific ayah?
But don't make it a primary key, since it is not unique.
Default value `0` also doesn't make any sense.

quran/models.py
```diff
@@ -0,0 +1,18 @@
...
+    text = models.TextField()
```
al-muammar, line 15: Define `set_text` and `get_text` methods, in order to think about this field as an array and don't bother yourself with its internal representation in the database.

helpers/ayah/convert_quran_words_ar_csv.py
```diff
@@ -0,0 +1,58 @@
...
+    '''Form import ready ayah list json from structured dict'''
+    list_of_ayahs = []
+    for surah, ayahs in surah_dict.items():
+        for ayah, words in ayahs.items():
+            element = {
+                "model": "quran.Ayah",
+                "fields": {
+                    "surah": surah,
+                    "ayah": ayah,
+                    "text": words
```
al-muammar: I thought it should be something like `','.join(words)`. Otherwise, you will store square brackets and this is probably not what you intended.

al-muammar (changes requested):
Way better this time. You are very close, brother!
After updating models, don't forget to update the scripts!

helpers/ayah/convert_quran_words_ar_csv.py
govza: Sure, more pythonic I'd say

quran/models.py
govza: Why? Isn't this constant structure?

quran/models.py
govza: I added ayah field as identifier of ayah number.

quran/models.py
al-muammar: Can you define your setter and getter? That would be awesome. Look for example [here](https://stackoverflow.com/questions/22340258/django-list-field-in-model). But don't copy and paste.
Your setter should join your words with a comma, your getter should just split by comma. That's all.

al-muammar (approved):
Accepting the PR in order not to block it, but ideally you need to implement one small thing.
Nice job ;)

PR #2, opened 2018-10-05
Выводим аяты

safbek:


jscode.js
```diff
@@ -0,0 +1,22 @@
+var words = ['It', 'is', 'awfully', 'hard'];
```
al-muammar: Use `let`, because you don't want to change these words in the future. It is an additional check that will help you to avoid bugs in the future.

jscode.js
```diff
@@ -0,0 +1,22 @@
+var words = ['It', 'is', 'awfully', 'hard'];
+var times = [1000, 2000, 4000, 6000];
```
al-muammar: The same here, use `let`.

jscode.js
```diff
@@ -0,0 +1,22 @@
+var words = ['It', 'is', 'awfully', 'hard'];
+var times = [1000, 2000, 4000, 6000];
+
+function showAyah(words,times){
```
al-muammar: Please, format correctly.
```js
function showAyah(words, times) {
```

jscode.js
```diff
@@ -0,0 +1,22 @@
+var words = ['It', 'is', 'awfully', 'hard'];
+var times = [1000, 2000, 4000, 6000];
+
+function showAyah(words,times){
+
+	var htmlElement = document.querySelector(".ayahWords");
+	let wordCount = words.length;
```
al-muammar: Suppose, you made an error somewhere and passed incorrect `words` and `times` to the function. You want to catch this error as soon as possible. Add a check, that if the lengths of these two arrays different, then throw an exception. You can do it as follows:
```js
if (words.length !== times.length) {
    throw "Words and times lengths must be equal!";
}
```

jscode.js
```diff
@@ -0,0 +1,22 @@
+var words = ['It', 'is', 'awfully', 'hard'];
+var times = [1000, 2000, 4000, 6000];
+
+function showAyah(words,times){
+
+	var htmlElement = document.querySelector(".ayahWords");
+	let wordCount = words.length;
+
+	for(var currentIndex = 0; currentIndex < wordCount; currentIndex++) {
```
al-muammar: Rename `currentIndex` to just `index` or even `i`. Don't use extremely long names, if you have short alternatives.

jscode.js
```diff
@@ -0,0 +1,22 @@
...
+var times = [1000, 2000, 4000, 6000];
+
+function showAyah(words,times){
+
+	var htmlElement = document.querySelector(".ayahWords");
+	let wordCount = words.length;
+
+	for(var currentIndex = 0; currentIndex < wordCount; currentIndex++) {
+		
+		(function(index) {
```
al-muammar: This function doesn't add any value, remove it. I will show you an example. This is what you wrote:
```js
var n = 0;
function plus(x) {
    n += x;
}
for (var i = 0; i < 10; ++i) {
    plus(x);
}
console.log(n);
```
You don't really need this function. The code above is equivalent to:
```js
var n = 0;
for (var i = 0; i < 10; ++i) {
    n += i;
}
console.log(n);
```

jscode.js
```diff
@@ -0,0 +1,22 @@
...
+			let word = words[index];//это необходимо?
```
al-muammar: What is necessary? You can use `words[index]` and `times[index]` directly in `setTimeout` function.

al-muammar (changes requested):
Not so bad for your first commit!

jscode.js
al-muammar: Please, correct the formatting. You have to put a space after the comma.

jscode.js
```diff
@@ -0,0 +1,20 @@
+let words = ['It', 'is', 'awfully', 'hard'];
+let times = [1000, 2000, 4000, 6000];
+
+function showAyah(words,times){
+	let htmlElement = document.querySelector(".ayahWords");
+	let wordCount = words.length;
+	if(words.length !== times.length){
```
al-muammar: Format correctly here too.

jscode.js
```diff
@@ -0,0 +1,20 @@
+let words = ['It', 'is', 'awfully', 'hard'];
+let times = [1000, 2000, 4000, 6000];
+
+function showAyah(words,times){
+	let htmlElement = document.querySelector(".ayahWords");
+	let wordCount = words.length;
+	if(words.length !== times.length){
+		throw('Oops!');
```
al-muammar: Please, write some informative exception. Something like:
```js
throw(`The lengths of the arrays don't match: ${words.length} vs ${times.length}`)
```

jscode.js
```diff
@@ -0,0 +1,20 @@
...
+	for(let i = 0; i < wordCount; i++) {
+		let word = words[i];
+		let time = times[i];
+		setTimeout(function() {
+			htmlElement.innerHTML += " " + word;
+		}, time);
+	}
+
+}
+showAyah(words,times);
```
al-muammar: Add a single empty line after this one.

al-muammar:
You need to correct a few small formatting issues.
Move your code to `static/js/the_name_of_your_file.js`.

al-muammar (changes requested):
Wrote the comment separately.

static/js/jscode.js
```diff
@@ -0,0 +1,21 @@
+let words = ['It', 'is', 'awfully', 'hard'];
+let times = [1000, 2000, 4000, 6000];
+
+function showAyah(words, times) {
+	let htmlElement = document.querySelector(".ayahWords");
+	let wordCount = words.length;
```
al-muammar: Set up the identation everywhere to be 4 spaces instead of 8.

static/js/jscode.js
```diff
@@ -0,0 +1,21 @@
+let words = ['It', 'is', 'awfully', 'hard'];
+let times = [1000, 2000, 4000, 6000];
+
+function showAyah(words, times) {
+	let htmlElement = document.querySelector(".ayahWords");
+	let wordCount = words.length;
+	if(words.length !== times.length){
+		throw(`The lengths of the arrays don't match: ${words.length} vs ${times.length}`);
```
al-muammar: This line seems too long.

al-muammar (approved):
Some nits, otherwise good enough.
Great job!

PR #4, opened 2018-10-17
Separate folder for backend

govza:
Refactor, move backend to separate Folder.


al-muammar:
I like the fact that you separated your PRs, but I don't like this PR. Why do we need a separate folder? In any case, the frontend will live somewhere inside `static/` or `templates/`. There is no reason to create more folders.

govza:
> 
> 
> I like the fact that you separated your PRs, but I don't like this PR. Why do we need a separate folder? In any case, the frontend will live somewhere inside `static/` or `templates/`. There is no reason to create more folders.

I was thinking about separate folder for database, backend and frontend. Easy for testing, deployment in different docker images, logically divided for teams. Have a look to structure of https://github.com/govza/dockerized-django

rguliev:
Well, we've decided to use regular JavaScript for now, so I agree with @Jihadik and I also see no reason for extra complication as far as we just use regular JS (+ jQuery maybe). I would focus on v1.0 where we just use python, django, javascript (+jQuery maybe). Then, we can make it more advanced (with Vue, Docker and other stuff) later.

govza:
Actually I find it not very convenient to do work twice, first with JQuery than with Vue

rguliev:
About "work twice". Do not forget that we do it mostly for practicing/learning. So, it is rather "learning step by step (from simple to more complicated)", than "working twice".
About jQuery. Actually, we would like to avoid jQuery too. I just thought that we might need it to make ajax requests. So it does not make things more complicated.

al-muammar:
@govza, currently, we have nothing. When you will set up a working prototype with a minimal functionality, we can start thinking about more complicated parts.
1. Don't confuse two things: **modeling your project and modeling your code**. You should think about modeling your project and make it extensible for the addition of new features. But that doesn't mean that you have to add all of these features straight away and try to model the code from the beginning. Start with 1 feature, make it reliable, move forward. Add another feature, make it reliable, move forward. And so on. If you add 10 features and none of them works, then you accomplished nothing.
Actually, one more thing. If you overcomplicate your code, it slows you down. You have to think about all future dependencies, that are not in place. That's the waste of your time.
2. **Be humble and learn**. More experienced people trying to teach you best practices. We are coming to the same issue again and again.
3. **Really?** If you consider `mv my-project backend/my-project` a big change that you don't want to do twice, then I have bad news for you xD

govza:
I think there still good reasons for this PR so I reopen it.

1. I see what you mean, I think we need separate also apps for quran and recitations, to make project more extensible. But for this PR is more about structure than feature.
2. I agre, as programmers we learning every day, but humble is not the right word for peoples relations. I prefer be rational and intelligent instead.
3. This is not only mv, there also docker builders and travis configs as well as development perspectives - testing, etc.

al-muammar:
1. About separation. No, no and no. You don't have a working prototype, you can't separate from nothing. Make the prototype first and then start separating things.
2. These 3: humbleness, rationality and intelligence do not contradict. Homework: think about it and understand why.
3. Forget about docker and testing. You have nothing to run and nothing to test. You are creating obstacles and complications for yourself. You are slowing yourself down.

PR #6, opened 2018-11-07
Play surah

safbek:


static/js/mp3.js
```diff
@@ -0,0 +1,22 @@
+function leftPad (str, max) {
+  str = str.toString();
+  return str.length < max ? leftPad("0" + str, max) : str;
```
al-muammar, line 3: I'm 100% sure, that it is not your code :)
It's a pretty funny recursive implementation. I would make just a loop.

static/js/mp3.js
```diff
@@ -0,0 +1,22 @@
+function leftPad (str, max) {
```
al-muammar, line 1: Format the whole code, please.

al-muammar (changes requested):
1. Formatting, formatting and formatting.
2. Let's split your code into 3 functions:
```js
function generateUrl(reciter, surah, ayah)
function playAyah(reciter, surah, ayah)
function playSurah(reciter, surah)
```
The first one will generate a right url, something like `http://www.mywebsite.com/media/<reciter>/<surah>/<ayah>`
The second one for a given url will play the audio.
The third one will iterate through all ayahs in a surah and play all of them.

PR #7, opened 2018-11-08
Work with Reciter and Recitation models

mannpy:
Changing models, migration files. Adding new Reciter and Recitation models.


quran/models.py
```diff
@@ -1,19 +1,29 @@
...
     name = models.CharField(max_length=20)
     total_ayahs = models.PositiveIntegerField()
 
+    def __str__(self):
+        return self.name
+
 
 class Ayah(models.Model):
-    '''General information about Ayah'''
+
```
al-muammar: Usually, you don't put an empty line here.

quran/models.py
```diff
@@ -1,19 +1,29 @@
...
 class Surah(models.Model):
-    '''General information about Surahs'''
+
+    ''' General information about Surah '''
+
     number = models.PositiveIntegerField(primary_key=True)
     name = models.CharField(max_length=20)
     total_ayahs = models.PositiveIntegerField()
 
+    def __str__(self):
```
al-muammar, line 15: Depends on where are you going to use this function, but I would prefer to see a surah number in the brackets too.

quran/migrations/0002_load_data.py
```diff
@@ -4,34 +4,35 @@
...
+        fixture_file = path.join(self.fixture_dir, self.fixture_data)
+        call_command('loaddata', fixture_file)
 
 
-def load_ayahs(apps, schema_editor):
-    fixture_file = path.join(fixture_dir, fixture_ayahs)
-    call_command('loaddata', fixture_file)
+class UnloadData:
 
+    """ Deleting all entries for this model """
```
al-muammar:
```suggestion
    """Class for deleting all entries from a given (app, model) pair."""
```

quran/migrations/0002_load_data.py
```diff
@@ -4,34 +4,35 @@
...
 fixture_surahs = 'surah_list.json'
 fixture_ayahs = 'ayah_list.json'
 
 
-def load_surahs(apps, schema_editor):
-    fixture_file = path.join(fixture_dir, fixture_surahs)
-    call_command('loaddata', fixture_file)
+class LoadData:
 
+    """ Load fixture files """
```
al-muammar: Expand the description here, please.

al-muammar (changes requested):
LGTM except few small nits.

al-muammar:
English:
Reciter, Recitation, TimeCode classes **were** added. Fixture files generated

@mannpy, can you add some description about what is happening here and how were you able to add 750k lines?

quran/models.py
```diff
@@ -1,19 +1,67 @@
...
+        related_name='recitations')
+
+    def __str__(self):
+        return "%s:(%s)" % (self.reciter, self.url_mask)
+
+
+class TimeCode(models.Model):
+    '''Model representing a timecodes of the recitation ayahs'''
+    ayah_id = models.PositiveIntegerField()
+    segments = models.TextField()
```
al-muammar: Move it to `Recitation` and delete this model.
Call it `time_segments`.

quran/models.py
```diff
@@ -1,19 +1,67 @@
...
+        Reciter,
+        on_delete=models.CASCADE,
+        related_name='recitations')
+
+    def __str__(self):
+        return "%s:(%s)" % (self.reciter, self.url_mask)
+
+
+class TimeCode(models.Model):
+    '''Model representing a timecodes of the recitation ayahs'''
```
al-muammar: Time code**s** are plural, so you can't say **a** time codes. You should use no prepositions.

quran/models.py
```diff
@@ -1,19 +1,67 @@
...
+            # Newly created object, so set slug
+            self.slug = slugify(self.name)
+        super(Reciter, self).save(*args, **kwargs)
+
+    def __str__(self):
+        return self.name
+
+
+class Recitation(models.Model):
+    '''Model representing a recitation of the reciter'''
```
al-muammar: When you are writing a documentation, try to explain it to a random observer.
That's ok to write them like this now, but in the future try to be more descriptive.

For example:
```
This model represents a recitation of an ayah from the Qur'an.
Each recitation has an audio file and time segments which hold
an information about times when each word in this audio file was pronounced.
```

quran/models.py
```diff
@@ -1,19 +1,67 @@
...
+            self.surah.name, self.surah.number, self.number)
+
+
+class Reciter(models.Model):
+    '''Model representing a reciter'''
+    name = models.CharField(max_length=100)
+    slug = models.SlugField(unique=True)
+
+    def save(self, *args, **kwargs):
+        if not self.id:
```
al-muammar: https://stackoverflow.com/questions/2165865/django-queries-id-vs-pk

al-muammar (changes requested):
Few comments here and there, but overall LGTM.

quran/models.py
```diff
@@ -1,19 +1,88 @@
 '''Models describing abstractions used at Quran app'''
-from django.db import migrations, models
+from os import path
```
al-muammar: Please, don't do it, `os.path` is short enough.
```suggestion
import os
```

quran/migrations/0001_initial.py
```diff
@@ -1,7 +1,8 @@
-# Generated by Django 2.1 on 2018-09-28 19:51
+# Generated by Django 2.1 on 2018-11-14 04:56
 
 from django.db import migrations, models
 import django.db.models.deletion
```
al-muammar: Remove this `import`. It doesn't add any value.

quran/migrations/0001_initial.py
```diff
@@ -1,7 +1,8 @@
...
+import quran.models
```
al-muammar, line 4: I prefer to structure my imports as follows:
```py
# Native python imports
import os
import logging

# Third party libraries imports
from django.db import migrations, models

# Your code imports
import quran.models
```

quran/models.py
```diff
@@ -1,19 +1,88 @@
...
     name = models.CharField(max_length=20)
     total_ayahs = models.PositiveIntegerField()
 
+    def __str__(self):
+        return "%s:(%s)" % (self.name, self.number)
+
 
 class Ayah(models.Model):
-    '''General information about Ayah'''
+    '''Model representing a ayah from the Qur’an.'''
```
al-muammar:
```suggestion
    '''Model representing an ayah from the Qur’an.'''
```

quran/models.py
```diff
@@ -1,19 +1,88 @@
...
+    def __str__(self):
+        return "%s: (%s:%s)" % (
+            self.surah.name, self.surah.number, self.number)
+
+
+class Reciter(models.Model):
+    '''Model representing a reciter'''
+    name = models.CharField(max_length=100)
+    quality = models.CharField(max_length=20, blank=True)
+    style = models.CharField(max_length=100, blank=True)
```
al-muammar: Do we really need 100 symbols for this field?

quran/models.py
```diff
@@ -1,19 +1,88 @@
...
+    slug = models.SlugField()
```
al-muammar: Can you add a `help`?

quran/models.py
```diff
@@ -1,19 +1,88 @@
...
+
+    def __str__(self):
+        return "%s: (%s:%s)" % (
+            self.surah.name, self.surah.number, self.number)
+
+
+class Reciter(models.Model):
+    '''Model representing a reciter'''
+    name = models.CharField(max_length=100)
+    quality = models.CharField(max_length=20, blank=True)
```
al-muammar: Can you add a small `help` with what kind of strings we expect for `quality`?
Add a validator for it, please.

quran/models.py
```diff
@@ -1,19 +1,88 @@
...
+    style = models.CharField(max_length=100, blank=True)
+    slug = models.SlugField()
+
+    def save(self, *args, **kwargs):
+        self.quality = self.quality.replace('/', 'p').replace(' ', '')
```
al-muammar: What kind of magic is happening here? You should add a comment at least. But ideally we can make it simpler.

quran/models.py
```diff
@@ -1,19 +1,88 @@
...
+        self.slug = slugify(self.name)
+        super(Reciter, self).save(*args, **kwargs)
+
+    def __str__(self):
+        return self.name
```
al-muammar: I would add all information, i.e. name + quality + style.

quran/models.py
```diff
@@ -1,19 +1,88 @@
...
+
+
+def audio_directory_path(instance, filename):
```
al-muammar:
```suggestion
def audio_directory_path(recitation, filename):
```

quran/models.py
```diff
@@ -1,19 +1,88 @@
...
+    '''
+    the function generates path for
```
al-muammar:
```suggestion
    Return a path where audio file for a given recitation will be uploaded.
```

quran/models.py
```diff
@@ -1,19 +1,88 @@
...
+
+
+class Recitation(models.Model):
+    '''
+    This model represents a recitation of an ayah from the Qur'an.
+    Each recitation has an audio file and time segments which hold
+    an information about times when each word in this audio file
+    was pronounced.
+    '''
+    ayah_id = models.PositiveIntegerField()
```
al-muammar: You are reinventing a `ForeignKey` here. Please, don't do it :)

quran/models.py
```diff
@@ -1,19 +1,88 @@
...
+        return self.name
+
+
+def audio_directory_path(instance, filename):
+    '''
+    the function generates path for
+    uploading audio file of ayahs to
+    MEDIA_ROOT/<reciter-slug>/<style>/<quality>/<surah-id>/<ayah-id>.mp3
+    '''
+    ayah = Ayah.objects.get(pk=instance.ayah_id)
```
al-muammar: The code is pretty complicated, follow the instructions for `Recitation` model and update the code accordingly.

quran/models.py
```diff
@@ -1,19 +1,88 @@
...
+
+class Recitation(models.Model):
+    '''
+    This model represents a recitation of an ayah from the Qur'an.
+    Each recitation has an audio file and time segments which hold
+    an information about times when each word in this audio file
+    was pronounced.
+    '''
+    ayah_id = models.PositiveIntegerField()
+    segments = models.TextField()
```
al-muammar: According to https://stackoverflow.com/questions/22340258/django-list-field-in-model you can redefine your own getter and setter for the field. I never tried it, but it looks very promising. Can you write your custom `getter` and `setter` for this field?
Test it and show that it works fine.

al-muammar (changes requested):
Quite a few changes to make, but you are on a right track!

quran/models.py
mannpy: OK

quran/migrations/0001_initial.py
mannpy: We can't remove it, because it' using. Look at lines 63, 70

quran/migrations/0001_initial.py
mannpy: OK

quran/models.py
mannpy: Thanks

quran/models.py
mannpy: No)

quran/models.py
mannpy: OK)

quran/models.py
mannpy: OK. But I should redefine model's clean function to validate during save model.

quran/models.py
mannpy: How we can make it simpler?

quran/models.py
mannpy: Ok

quran/models.py
mannpy: Good

quran/models.py
mannpy: OK)

quran/models.py
mannpy: Ok, but we should give to setter function list of segments. It will look like after dumping
'[[0,1,180,1230],[1,2,1240,2160],[2,3,2170,3040],[3,4,3050,4380],[4,5,4390,5410],[42,43,59530,62890]]'
And we should make something like [i[2:] for i in segments] before giving to setter function

quran/migrations/0001_initial.py
al-muammar: Use the previous line of import and refactor lines 63 and 70.

quran/models.py
```diff
@@ -1,19 +1,129 @@
...
+
+    def __str__(self):
+        return "%s: (%s:%s)" % (
+            self.surah.name, self.surah.number, self.number)
+
+
+class Reciter(models.Model):
+    '''Model representing a reciter'''
+    name = models.CharField(max_length=100)
+    quality = models.CharField(
```
al-muammar: Let’s call this field `bitrate` and store it as an integer.

quran/models.py
al-muammar: This is the wrong place to do these transformations. You should make bitrate an integer field and make all transformations and extractions in your forms.

quran/models.py
```diff
@@ -1,19 +1,129 @@
...
+        containing only letters and hyphens. \
+        It's filled automatically during saving.")
+
+    def clean(self, *args, **kwargs):
+        # replacing slashes and empty symbols to avoid creating subfolders
+        # while audio file upload to media
+        self.quality = self.quality.replace('/', 'p').replace(' ', '')
+        # validating bitrate quality matching to string like '192kbps'
+        bitrate_validator = RegexValidator(
+            regex=r'^[\d]{1,3}\w{2}[p|\/]\w$',
```
al-muammar: Nice approach!
I would make it more precise, though. Add for sure we need testing around this method!

quran/models.py
```diff
@@ -1,19 +1,129 @@
...
+        self.slug = slugify(self.name)
+        # running redefined full_clean to validate bitrate quality
+        self.full_clean()
+        super(Reciter, self).save(*args, **kwargs)
+
+    def __str__(self):
+        return "%s: %s(%s)" % (self.name, self.style, self.quality)
+
+
+def audio_directory_path(recitation, filename):
```
al-muammar, line 60: I don’t like how you wrote this function:
1. Don’t use ayah-pk, ayah-number is unique in that context
2. Don’t use `os.path.join` in intermediate steps, use it in a final step only. Currently, you are abusing this function :)

quran/models.py
```diff
@@ -1,19 +1,129 @@
...
+    reciter = models.ForeignKey(
+        Reciter,
+        on_delete=models.CASCADE,
+        related_name='recitations')
+    segments = models.TextField()
+    audio = models.FileField(upload_to=audio_directory_path)
+
+    def set_segments(self, segments_list):
+        '''set ayah's segments list to field of segments as a string.'''
+        self.segments = json.dumps(segments_list, separators=(',', ':'))
```
al-muammar: Don’t use json dump. We don’t really need json format here.

al-muammar (changes requested):
I like the direction where everything goes. I think you should add a simple form view to this PR. You will see those problems that I mentioned.

quran/models.py
```diff
@@ -1,19 +1,107 @@
...
+
+    def __str__(self):
+        return f'{self.surah.name}: ({self.surah.number}:{self.number})'
+
+
+class Reciter(models.Model):
+    '''Model representing a reciter'''
+    name = models.CharField(max_length=100)
+    bitrate = models.PositiveIntegerField(
+        blank=True,
```
al-muammar: https://stackoverflow.com/questions/8609192/differentiate-null-true-blank-true-in-django
It really depends on how we are going to use it. I think we should add `null=True`, because sometimes we won't know the bitrate of an audio.

quran/models.py
```diff
@@ -1,19 +1,107 @@
...
+        help_text='This field contain a bitrate of audio file')
```
al-muammar:
```suggestion
        help_text='Bitrate of an audio file')
```

quran/models.py
```diff
@@ -1,19 +1,107 @@
...
+    style = models.CharField(
+        max_length=20, blank=True,
+        help_text='This filed describe style \
```
al-muammar:
```suggestion
        help_text='Qur'an reading style'
```

quran/models.py
```diff
@@ -1,19 +1,107 @@
...
+        super(Reciter, self).save(*args, **kwargs)
+
+    def __str__(self):
+        return f'{self.name}: {self.style}({self.bitrate}kb/s)'
+
+
+def audio_directory_path(recitation, filename):
+    '''
+    Return a path where audio file for a given recitation will be uploaded
+    MEDIA_ROOT/<reciter-slug>/<style>/<bitrate>/<surah-pk>/<ayah-pk>.mp3
```
al-muammar:
```suggestion
    MEDIA_ROOT/<reciter-slug>/<style>/<bitrate>/<surah-number>/<ayah-number>.mp3
```

quran/models.py
```diff
@@ -1,19 +1,107 @@
...
+        '''
+        set ayah's segments list containing tuples
+        of timecodes pair of beginning and ending
+        of the word to field of segments as a string.
+        '''
+        self.segments = ','.join(
+            (f'{segment_pair[0]}:{segment_pair[1]}'
+                for segment_pair in segments_list))
+
+    @property
```
al-muammar: Are you sure that it should be a property? Have you tried creating, saving and loading `Recitation` objects?
We definitely need to write a unit test for it!

al-muammar (changes requested):
Nicely done!
Small comments:
1. Change the name and the description of the PR. Add `Test Plan` somewhere in the description and describe (but better is to show) how you tested your code.
2. There is no need to write in `help_text` "this field ..."

Substantial changes that you need to do:
1. Add unit tests around getter and setter in `Recitation`
2. Add a simple view that is populating these models. It should be able to take a meta information (Reciter's name, e.t.c) and url form Qur'an Academy and populate a database with data.

quran/fields.py
```diff
@@ -0,0 +1,44 @@
+from django.core import exceptions
+from django.db import models
+
+
+class SegmentsField(models.TextField):
+    description = "A custom field to save list of timecodes in DB as a string"
```
al-muammar: Look at examples how do Django developers define descriptions: https://docs.djangoproject.com/en/2.1/_modules/django/db/models/fields/#TextField

quran/fields.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def to_python(self, segments):
```
al-muammar: If you look at `CharField` docs you will see that they always call it `value`:
```suggestion
    def to_python(self, value):
```

quran/fields.py
```diff
@@ -0,0 +1,44 @@
...
+        if isinstance(segments, list) or segments is None or segments == '':
```
al-muammar: This is wrong. Let's look at it step by step:
1. There are not enough checks here, let's say you will think that list of `[None, 25, (389,-1)]` is a right value. Let's omit this for now and pretend that it should always be string or None.
2. `segments is None` is ok, everything is right here.
3. `segments == ''` -- this is wrong. You always want to return a list of tuples, but in this case you will return an empty string! It is problematic. You should return an empty list in this case. **Cover it with a test**.

quran/fields.py
```diff
@@ -0,0 +1,44 @@
...
+            return segments
+        try:
+            return [
```
al-muammar: This is right, but slightly convoluted, separate it into two lines:
```py
string_segments = segments.split(',')
return [(int(l), int(r)) for l, r in segment.split(':') for segment in string_segments]
```

quran/fields.py
```diff
@@ -0,0 +1,44 @@
...
+from django.db import models
+
+
+class SegmentsField(models.TextField):
+    description = "A custom field to save list of timecodes in DB as a string"
+
+    def to_python(self, segments):
+        if isinstance(segments, list) or segments is None or segments == '':
+            return segments
+        try:
```
al-muammar: Write a comment describing what you are trying to do here.

quran/fields.py
```diff
@@ -0,0 +1,44 @@
...
+            return [
+                tuple(int(timecode) for timecode in timecodes.split(':'))
+                for timecodes in segments.split(',')]
+        except (TypeError, ValueError):
```
al-muammar: `IndexError` could also happen here.

quran/fields.py
```diff
@@ -0,0 +1,44 @@
...
+            raise exceptions.ValidationError(
+                message='Invalid string of segments',
+                code='invalid',
+            )
+
+    def from_db_value(self, value, expression, connection):
+        return self.to_python(value)
+
+    def get_prep_value(self, segments_list):
```
al-muammar:
```suggestion
    def get_prep_value(self, value):
```

al-muammar: You should also call as a first line you parent's `get_prep_value` like this:
`super().get_prep_value()`

quran/fields.py
```diff
@@ -0,0 +1,44 @@
...
+        if segments_list is None or segments_list == '':
```
al-muammar: Usually, people try converting everything `to_python()` first and then make conversions with no checks.

quran/models.py
```diff
@@ -1,19 +1,97 @@
...
+        help_text="Short label for name, \
+        containing only letters and hyphens. \
+        It's filled automatically during saving.")
+
+    def save(self, *args, **kwargs):
+        # creating slug from name
+        self.slug = slugify(self.name)
+        super(Reciter, self).save(*args, **kwargs)
+
+    def __str__(self):
```
al-muammar, line 53: This is too complicated. Rewrite it in an easier way. Something like: `Name (params)` and if some of the params is absent, put `None` there. Example: `Mishari (style=fast, bitrate=None)`

quran/test_models.py
```diff
@@ -0,0 +1,105 @@
...
+        self.mishari.style = 'murattal'
+        self.mishari.bitrate = 128
+        self.assertEqual(
+            str(self.mishari), 'Mishari Rashid al-ʿAfasi-murattal-128kb/s')
+
+    def test_reciter_slug(self):
+        self.assertEqual(self.mishari.slug, 'mishari-rashid-al-afasi')
+
+
+class RecitationTestCase(TestCase):
```
al-muammar, line 41: You should write a separate test case for your `SegmentsField`. It's independent of your model in fact. Try to read some Django code and see how do they test their fields.

al-muammar (changes requested):
We are heading to the right direction! Many comments, but everything becomes better and better.

quran/fields.py
mannpy: But, we don't need the translation of this field. description = _("Text") - _ - it's gettext_lazy, and it's using for translation. Or are you mean that we should have a short description, like "Segments"?

quran/fields.py
mannpy: I don't understand why, but it doesn't work. It returns, that name 'segment' is not defined. Try it yourself.
```python
string_segments = '12:23,34:56,78:90'.split(',')
[(int(l), int(r)) for l, r in segment.split(':') for segment in string_segments]
```

mannpy: Only can to make:
```python
string_segments = value.split(',')
return [tuple(
      int(timecode) for timecode in timecodes.split(':'))
      for timecodes in string_segments]
```
But it's look more confusing

quran/fields.py
mannpy: It's written in the official documentation. They first make conversation with no check, and after that convert to python objects.
Look at https://docs.djangoproject.com/en/2.1/_modules/django/db/models/fields/#IntegerField. or https://docs.djangoproject.com/en/2.1/_modules/django/db/models/fields/#FloatField
In https://docs.djangoproject.com/en/2.1/_modules/django/db/models/fields/#TextField. they first converting `to_python()` because textfield is same string in DB, and in python. Therefore get_prep_value and to_python are same.

quran/fields.py
mannpy: OK

mannpy: There is some error If I add this line. Because `super.get_prep_value` runs `to_python`, that we don't need in our case. Because  `get_prep_value` is using to load data from python objects to DB string format, `to_python` from DB string to python objects. So I won't add it. Works fine without it.

quran/models.py
```diff
@@ -1,19 +1,95 @@
...
+    name = models.CharField(max_length=100)
+    bitrate = models.PositiveIntegerField(
+        blank=True,
+        null=True,
+        help_text="Bitrate of an audio file")
+    style = models.CharField(
+        max_length=20, blank=True,
+        help_text="Qur'an reading style")
+    slug = models.SlugField(
+        help_text="Short label for name, \
```
al-muammar: Use the following notation:
```py
"Short label for name, "
"containing only letters and hyphens. "
"It's filled automatically during saving."
```
The strings will be concatenated for you.

quran/models.py
```diff
@@ -1,19 +1,95 @@
...
+        containing only letters and hyphens. \
+        It's filled automatically during saving.")
+
+    def save(self, *args, **kwargs):
+        # creating slug from name
+        self.slug = slugify(self.name)
+        super(Reciter, self).save(*args, **kwargs)
+
+    def __str__(self):
+        return f'{self.name} ' + \
```
al-muammar: Install `black` python formatter and format this file. It will make everything way nicer.

quran/test_fields.py
```diff
@@ -0,0 +1,216 @@
...
+from django.db import connection
+from django.core import exceptions
+from django.core.files import File
+from django.conf import settings
+from django.test import TestCase, SimpleTestCase
+from .models import Ayah, Reciter, Recitation
+from .fields import SegmentsField
+
+
+class SegmentsFieldTestCase(TestCase):
```
al-muammar, line 7: Remove all models and simplify your tests. You wrote a pretty generic field that should be tested in isolation of any models, you don't really need them.

quran/fields.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+class SegmentsField(models.TextField):
+    description = "Segments"
+
+    def to_python(self, value):
+        if value is None:
+            return None
+        # if value is empty string return empty list
+        if value == '':
```
al-muammar: I think it is automatically covered in your `try` clause, so I would remove it.

quran/fields.py
```diff
@@ -0,0 +1,82 @@
...
+        # [(12, 15), (16, 19)] => "12:15,16:19"
+        return ','.join(
+            (f'{segment_pair[0]}:{segment_pair[1]}'
+                for segment_pair in value))
+
+    def value_to_string(self, obj):
+        value = self.value_from_object(obj)
+        return self.get_prep_value(value)
+
+    def _validate_tuples_containing_in_list(self, value):
```
al-muammar: Let's make it shorter
```suggestion
    def _validate_tuples(self, value):
```

quran/fields.py
```diff
@@ -0,0 +1,82 @@
...
+        for segments in value:
+            # if no tuples in value
+            if type(segments) != tuple:
+                raise ValidationError(
+                    message='List must contains tuples only',
+                    code='invalid')
+
+    def _validate_length_of_segments(self, value, message):
```
al-muammar: 1. Ditto
2. Remove message, as this function is checking lengths of given segments only, so there is no reason in providing additional message.
```suggestion
    def _validate_lengths(self, value):
```

quran/fields.py
```diff
@@ -0,0 +1,82 @@
...
+        return ','.join(
+            (f'{segment_pair[0]}:{segment_pair[1]}'
+                for segment_pair in value))
+
+    def value_to_string(self, obj):
+        value = self.value_from_object(obj)
+        return self.get_prep_value(value)
+
+    def _validate_tuples_containing_in_list(self, value):
+        for segments in value:
```
al-muammar: This code is fine, but you could rewrite it as follows:
```py
all_tuples = all(isinstance(segment, tuple) for segment in value)
if not all_tuples:
    ...
```

quran/fields.py
```diff
@@ -0,0 +1,82 @@
...
+            # if no tuples in value
+            if type(segments) != tuple:
+                raise ValidationError(
+                    message='List must contains tuples only',
+                    code='invalid')
+
+    def _validate_length_of_segments(self, value, message):
+        '''Check if length of tuples in a list equal two'''
```
al-muammar:
```suggestion
        '''Check if all tuple lengths equal two.'''
```

quran/fields.py
```diff
@@ -0,0 +1,82 @@
...
+        for segments in value:
```
al-muammar: Try to use `all` operator as was proposed above.

quran/fields.py
```diff
@@ -0,0 +1,82 @@
...
+from django.db import models
+
+
+class SegmentsField(models.TextField):
+    description = "Segments"
+
+    def to_python(self, value):
+        if value is None:
+            return None
+        # if value is empty string return empty list
```
al-muammar: This comment is too obvious, remove it.

quran/fields.py
```diff
@@ -0,0 +1,82 @@
...
+        if value == '':
+            return []
+        # if value is list
```
al-muammar: This comment is too obvious too.

quran/fields.py
```diff
@@ -0,0 +1,82 @@
...
+        if isinstance(value, list):
+            # if list is not empty
```
al-muammar: This is obvious too, remove.

quran/fields.py
```diff
@@ -0,0 +1,82 @@
...
+            if value:
```
al-muammar: Your functions work (or should work!) for empty lists too, so there is no reason for this check.

quran/fields.py
```diff
@@ -0,0 +1,82 @@
...
+                # validate list of tuples
+                self._validate_tuples_containing_in_list(value)
+                self._validate_length_of_segments(
+                    value, "Length of tuples must be equals 2")
+            return value
+        try:
+            # try to parse a string of segments from DB to a list of tuples
```
al-muammar: There is no DB in this function whatsoever ;)
```suggestion
            # try to convert a string of segments to a list of tuples
```

quran/fields.py
```diff
@@ -0,0 +1,82 @@
...
+            # "12:15,16:19" => [(12, 15), (16, 19)]
+            segments_list = [tuple(
+                int(timecode) for timecode in timecodes.split(':'))
+                for timecodes in value.split(',')]
+            # validate list of tuples
+            self._validate_length_of_segments(
+                segments_list, 'Invalid string of segments')
+            return segments_list
+
```
al-muammar: Remove this empty line. One more time, install python `black` formatter.

quran/fields.py
```diff
@@ -0,0 +1,82 @@
...
+            if value:
+                # validate list of tuples
+                self._validate_tuples_containing_in_list(value)
+                self._validate_length_of_segments(
+                    value, "Length of tuples must be equals 2")
+            return value
+        try:
+            # try to parse a string of segments from DB to a list of tuples
+            # "12:15,16:19" => [(12, 15), (16, 19)]
+            segments_list = [tuple(
```
al-muammar: Don't use a plural noun as an adjective.
```suggestion
            segment_list = [tuple(
```

quran/fields.py
```diff
@@ -0,0 +1,82 @@
...
+                int(timecode) for timecode in timecodes.split(':'))
```
al-muammar: You are implementing `SegmentsFiled`, which is independent of time codes. You should think about it in abstraction and only when you will use it in your model, you can add some additional meaning to it.

quran/fields.py
```diff
@@ -0,0 +1,82 @@
...
+        except (TypeError, ValueError, IndexError):
+            raise ValidationError(
+                message='Invalid string of segments',
+                code='invalid')
+
+    def from_db_value(self, value, expression, connection):
+        return self.to_python(value)
+
+    def get_prep_value(self, value):
+        # save to DB None or empty string
```
al-muammar: Add this magical line just in case:
```py
value = super().get_prep_value(value)
```
In general, this function can be simplified:
```py
value = super().get_prep_value(value)
value = self.to_python(value)  # <- this function has all checks incapsulated
return ','.join(f'{segment[0]}:{segment[1]}' for segment in value)
```

quran/test_fields.py
```diff
@@ -0,0 +1,216 @@
...
+        self.test_dir = tempfile.mkdtemp()
+        self._original_media_root = settings.MEDIA_ROOT
+        settings.MEDIA_ROOT = self.test_dir
+        # get the 93rd ayah from the surah Al-Baqara
+        self.ayah_2_93 = Ayah.objects.get(pk=100)
+        # create a temporary audio
+        file = tempfile.NamedTemporaryFile(suffix='.mp3')
+        self.audio_ayah = File(file, name=file.name)
+        # create a reciter
+        self.husari = Reciter.objects.create(
```
al-muammar: Just by looking at this line I understood that your tests are wrong. You don't need any models to test your fields. Read few tests how other django fields are tested: https://github.com/django/django/tree/master/tests/model_fields

What you really need to test is your few methods converting from strings to python objects and vice versa.

quran/test_fields.py
```diff
@@ -0,0 +1,216 @@
...
+        self.assertIsInstance(recitation.segments, list)
+        self.assertEqual(recitation.segments, [(1, 2), (3, 4), (5, 6)])
+        # Refetch for save
+        recitation = Recitation.objects.defer("segments").get(pk=recitation.pk)
+        recitation.save()
+        recitation = Recitation.objects.get(pk=recitation.pk)
+        self.assertIsInstance(recitation.segments, list)
+        self.assertEqual(recitation.segments, [(1, 2), (3, 4), (5, 6)])
+
+    def test_recitation_with_empty_segments(self):
```
al-muammar: You don't need recitation here. Remove all unnecessary stuff.

quran/test_fields.py
```diff
@@ -0,0 +1,216 @@
...
+    def test_refresh_segments(self):
+        recitation = Recitation.objects.create(
+            ayah=self.ayah_2_93, segments=[(12, 15), (15, 19), (20, 26)],
+            reciter=self.husari, audio=self.audio_ayah)
+        recitation.refresh_from_db(fields=['segments'])
+        self.assertIsInstance(recitation.segments, list)
+        self.assertEqual(
+            recitation.segments, [(12, 15), (15, 19), (20, 26)])
+
+    def test_defer(self):
```
al-muammar: It's a recitation test, remove.

quran/test_fields.py
```diff
@@ -0,0 +1,216 @@
...
+        # create a reciter
+        self.husari = Reciter.objects.create(
+            name="Mahmud Kh. al-Husary")
+
+    def tearDown(self):
+        # Delete a temporary directory
+        shutil.rmtree(self.test_dir)
+        settings.MEDIA_ROOT = self._original_media_root
+
+    def test_refresh_segments(self):
```
al-muammar: It's a recitation test, remove.

quran/test_fields.py
```diff
@@ -0,0 +1,216 @@
...
+    def test_recitation_with_empty_segments(self):
+        recitation = Recitation.objects.create(
+            ayah=self.ayah_2_93, segments=[],
+            reciter=self.husari, audio=self.audio_ayah)
+        recitation.save()
+        recitation = Recitation.objects.get(pk=recitation.pk)
+        self.assertIsInstance(recitation.segments, list)
+        self.assertEqual(recitation.segments, [])
+
+    def test_recitation_with_empty_string_segments(self):
```
al-muammar: You don't need recitation here. Remove all unnecessary stuff.

quran/test_fields.py
```diff
@@ -0,0 +1,216 @@
...
+        recitation = Recitation.objects.create(
+            ayah=self.ayah_2_93, segments="",
+            reciter=self.husari, audio=self.audio_ayah)
+        recitation.save()
+        recitation = Recitation.objects.get(pk=recitation.pk)
+        self.assertIsInstance(recitation.segments, list)
+        self.assertEqual(recitation.segments, [])
+
+    def test_recitation_with_wrong_length_segments_more_than_two(self):
```
al-muammar: You don't need recitation here. Remove all unnecessary stuff.

quran/test_fields.py
```diff
@@ -0,0 +1,216 @@
...
+        try:
+            Recitation.objects.create(
+                ayah=self.ayah_2_93, segments=[(12, 15), (15, 90, 100)],
+                reciter=self.husari, audio=self.audio_ayah)
+        except exceptions.ValidationError as e:
+            self.assertEqual(
+                str(e), "['Length of tuples must be equals 2']")
+
+    def test_recitation_with_wrong_length_segments_less_than_two(self):
```
al-muammar: You don't need recitation here. Remove all unnecessary stuff.

quran/test_fields.py
```diff
@@ -0,0 +1,216 @@
...
+        with self.assertRaises(exceptions.ValidationError) as e:
+            Recitation.objects.create(
+                ayah=self.ayah_2_93, segments=[(12, 15), ()],
+                reciter=self.husari, audio=self.audio_ayah)
+        self.assertEqual(
+            "['Length of tuples must be equals 2']", str(e.exception))
+
+    def test_wrong_segments_not_containing_tuples(self):
```
al-muammar: You don't need recitation here. Remove all unnecessary stuff.

quran/test_fields.py
```diff
@@ -0,0 +1,216 @@
...
+        with self.assertRaises(exceptions.ValidationError) as e:
+            Recitation.objects.create(
+                ayah=self.ayah_2_93, segments=[12, 15, [15, 90, 100], None],
+                reciter=self.husari, audio=self.audio_ayah)
+        self.assertEqual(
+            "['List must contains tuples only']", str(e.exception))
+
+    def test_wrong_segments_with_no_list(self):
```
al-muammar: You don't need recitation here. Remove all unnecessary stuff.

quran/test_fields.py
```diff
@@ -0,0 +1,216 @@
...
+        with self.assertRaises(exceptions.ValidationError) as e:
+            Recitation.objects.create(
+                ayah=self.ayah_2_93, segments=(3, 5, 6),
+                reciter=self.husari, audio=self.audio_ayah)
+        self.assertEqual("['Segments must be a list']", str(e.exception))
+
+    def test_segments_get_prep_value(self):
```
al-muammar: This is a good test!

al-muammar (changes requested):
That's a good try, though you need some more experience in writing tests + separating the components. That's fine, it comes with experience :)

I'm still waiting a simple form view using these models from you :)

quran/fields.py
mannpy: No, I will have an error.

quran/fields.py
```diff
@@ -0,0 +1,63 @@
...
+            return value
+        try:
+            # try to convert a string of segments to a list of tuples
+            # "12:15,16:19" => [(12, 15), (16, 19)]
+            segment_list = [
+                tuple(int(segment) for segment in segments.split(":"))
+                for segments in value.split(",")
+            ]
+            # validate the obtained list of tuples
+            if not all((len(segment) == 2 for segment in segment_list)):
```
al-muammar: It is exactly `_validate_lengths` method. Use it, instead of implementing it one more time.

quran/fields.py
```diff
@@ -0,0 +1,63 @@
...
+class SegmentsField(models.TextField):
+    description = "Segments"
+
+    def to_python(self, value):
+        if value is None:
+            return None
+        if not value:
+            return []
+        if not isinstance(value, list) and not isinstance(value, str):
+            self._raise_validation_error("Segments must be a list")
```
al-muammar:
```suggestion
            self._raise_validation_error("Unable to decode anything other than list and string")
```

quran/fields.py
```diff
@@ -0,0 +1,63 @@
...
+        if isinstance(value, list):
```
al-muammar: At this point we know for sure that the value is either list or string. Let's check if it is a string first, convert it to list, and then make all validations. In this code you are doing validation twice, which is OK, but could be better.

quran/fields.py
```diff
@@ -0,0 +1,63 @@
...
+        if not all_tuples:
+            self._raise_validation_error("List must contains tuples only")
+
+    def _validate_lengths(self, value):
+        """Check if all tuple lengths equal two."""
+        all_equal_two = all((len(segment) == 2 for segment in value))
+        if not all_equal_two:
+            self._raise_validation_error("Length of tuples must be equals 2")
+
+    def _raise_validation_error(self, message):
```
al-muammar: Don't wrap one line with another. Raising an exception should be explicit.

quran/fields.py
```diff
@@ -0,0 +1,63 @@
...
+        """Check if only tuples contain in the list"""
+        all_tuples = all(isinstance(segment, tuple) for segment in value)
+        if not all_tuples:
+            self._raise_validation_error("List must contains tuples only")
+
+    def _validate_lengths(self, value):
+        """Check if all tuple lengths equal two."""
+        all_equal_two = all((len(segment) == 2 for segment in value))
+        if not all_equal_two:
+            self._raise_validation_error("Length of tuples must be equals 2")
```
al-muammar:
```suggestion
            self._raise_validation_error("Length of tuples must be equal to 2")
```

quran/fields.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def value_to_string(self, obj):
+        value = self.value_from_object(obj)
+        return self.get_prep_value(value) if value else ""
+
+    def _validate_tuples(self, value):
+        """Check if only tuples contain in the list"""
+        all_tuples = all(isinstance(segment, tuple) for segment in value)
+        if not all_tuples:
+            self._raise_validation_error("List must contains tuples only")
```
al-muammar:
```suggestion
            self._raise_validation_error("List must contain tuples only")
```

quran/test_fields.py
```diff
@@ -0,0 +1,107 @@
...
+        with self.assertRaisesMessage(ValidationError, expected_message):
+            self.field.get_prep_value([(3,), (5, 7)])
+        # wrong segments: length of tuples more than two
+        with self.assertRaisesMessage(ValidationError, expected_message):
+            self.field.get_prep_value([(3, 5), (5, 7, 9, 10)])
+
+    def test_segmentsfield_to_python(self):
+        self.assertIsNone(self.field.to_python(None))
+        self.assertEqual(self.field.to_python(""), [])
+        self.assertIsInstance(self.field.to_python("3:4,5:7"), list)
```
al-muammar: This is an extra check. Then next one is sufficient enough.

quran/test_fields.py
```diff
@@ -0,0 +1,107 @@
...
+        self.assertEqual(self.field.to_python("3:4,5:7"), [(3, 4), (5, 7)])
+        self.assertEqual(
+            self.field.to_python([(3, 4), (5, 7)]), [(3, 4), (5, 7)]
+        )
+        # wrong segments
+        expected_message = "['Invalid string of segments']"
+        with self.assertRaisesMessage(ValidationError, expected_message):
```
al-muammar, line 75: Can you add a test for `"a:b,c:d"`?

al-muammar (changes requested):
Few small comments, otherwise looks good to me!
I'm still waiting for a view :)

al-muammar:
Even though I have a number of comments to the PR, I'm merging it, since @mannpy won't have time to work on the project in the near future.

PR #11, opened 2019-02-09
Reciter and Recitations models refactored

govza:
Reciters and Recitations are moved to separate app. 
Use /admin to add Reciter and Recitations
(need to create superuser account first)

Notes: update Django and recreate db.sql
(some sql issues fixed in new version)


recite/apps.py
```diff
@@ -0,0 +1,5 @@
+from django.apps import AppConfig
+
+
+class ReciteConfig(AppConfig):
```
al-muammar, line 4: I have never seen anyone calling his class with a verb :)

recite/models.py
```diff
@@ -0,0 +1,53 @@
...
+        super(Reciter, self).save(*args, **kwargs)
+
+    def __str__(self):
+        return (
+            f"{self.name} "
+            f"(style={self.style or None}, bitrate={self.bitrate})"
+        )
+
+
+class Recitation(models.Model):
```
al-muammar, line 35: Where is a field for an audio file? Will it be added in a separate PR?

al-muammar (commented):
I'm pretty happy to accept the PR.
Can you add in the description the result of your tests? I.e. that they are passing.

recite/apps.py
govza: We have Reciters and Recitations models in this class. Didn't found noun to include both.

govza:
> Can you add in the description the result of your tests? I.e. that they are passing.

Tests are passing. I refactored them, he-he. May be next PR we need to implement travis CI

recite/admin.py
```diff
@@ -26,8 +26,10 @@ def has_delete_permission(self, request, obj=None):
 
 
 class ReciterForm(forms.ModelForm):
-    """Reciter Form has additional csv_file field of Recitations model"""
-    csv_file = forms.FileField()
+    """Reciter Form has additional segments_file field of Recitations model"""
+    segments_file = forms.FileField()
+    audio_files = forms.FileField(
```
al-muammar: I think, we decided that it should be uploaded in the form of zip archive, right?

recite/admin.py
```diff
@@ -45,33 +47,58 @@ class ReciterAdmin(admin.ModelAdmin):
...
+                file_ayah = file_mp3
+                break
+
+        if not file_ayah:
+            raise ValueError(
+                f"Not found Recitation audio file for ayah {ayah}")
+
+        return file_ayah
+
+    def handle_csv_data(self, segments_file, encoding):
```
al-muammar: Let's call it `read_segments_csv`?

al-muammar (approved):
I see some steady progress, barak Allahu fika, however, since we don't have quick review process and stacked PRs in GitHub are hard, you tend to put everything into on PR.
Let's merge it and refactor later if we need.

recite/models.py
govza: Added field for audio in this PR.

recite/admin.py
govza: Just saw your comment. I already updated PR with uploading zip, and deleting files as well, to finish the Model part so far.

recite/apps.py
al-muammar: So you took the common prefix xD
`RecitationConfig` sounds good :)

recite/models.py
```diff
@@ -0,0 +1,96 @@
...
+        Reciter, on_delete=models.CASCADE, related_name="recitations"
+    )
+    segments = SegmentsField()
+
+    def get_audio_directory(self, filename):
+        file_extension = os.path.splitext(os.path.basename(filename))[1]
+        return os.path.join(
+            "recite",
+            self.reciter.slug,
+            f"{self.ayah.surah.number:03d}",
```
al-muammar, line 56: whatever

recite/models.py
```diff
@@ -0,0 +1,96 @@
...
+            f"{self.ayah.number:03d}"
+            f"{file_extension}",
+        )
+
+    audio = models.FileField(
```
al-muammar, line 61: Formatting

recite/models.py
```diff
@@ -0,0 +1,96 @@
...
+        upload_to=get_audio_directory)
```
al-muammar, line 62: Does it work? Shouldn't it be:
```suggestion
        upload_to=self.get_audio_directory)
```

recite/models.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def __str__(self):
+        return f"{self.reciter}: ({self.ayah})"
+
+
+@receiver(models.signals.post_delete, sender=Reciter)
+def auto_delete_files_on_delete(sender, instance, **kwargs):
+    """
+    Deletes Recitation files from filesystem
+    when corresponding `Reciter` object is  deleted.
```
al-muammar, line 72:
```suggestion
    when the corresponding `Reciter` object is deleted.
```

recite/models.py
```diff
@@ -0,0 +1,96 @@
...
+        upload_to=get_audio_directory)
+
+    def __str__(self):
+        return f"{self.reciter}: ({self.ayah})"
+
+
+@receiver(models.signals.post_delete, sender=Reciter)
+def auto_delete_files_on_delete(sender, instance, **kwargs):
+    """
+    Deletes Recitation files from filesystem
```
al-muammar, line 71:
```suggestion
    Deletes audio files from the filesystem
```

recite/admin.py
```diff
@@ -0,0 +1,143 @@
...
+                recitation, _ = Recitation.objects.get_or_create(
+                    ayah=ayah,
+                    reciter=obj,
+                    segments=segments,
+                    audio=file_ayah
+                )
+
+        shutil.rmtree(temp_dir)
+
+    def get_file_paths(self, temp_dir):
```
al-muammar, line 90: Let's add here extension checking, so we return only audio files.

recite/admin.py
```diff
@@ -0,0 +1,143 @@
...
+            surah=surah_id,
+        )
+        return ayah
+
+    def get_file_audio_path(self, file_paths, ayah):
+        """Search in uploaded files for corresponding ayah"""
+
+        file_ayah_name_expected = str(ayah) + '.mp3'
+        file_ayah = None
+        for file_path in file_paths:
```
al-muammar, line 116: This will cost you around 36M operations. Let's make the complexity linear:
1. Read all rows from csv and add them to `segments[surah][ayah]` map
2. Read files one by one and extract the corresponding segment from the map

recite/admin.py
```diff
@@ -0,0 +1,143 @@
...
+                file_ayah = File(f)
+
+                recitation, _ = Recitation.objects.get_or_create(
+                    ayah=ayah,
+                    reciter=obj,
+                    segments=segments,
+                    audio=file_ayah
+                )
+
+        shutil.rmtree(temp_dir)
```
al-muammar, line 88: If somewhere above an exception is thrown, then this line won't be executed and you will leave this dir in your FS. Use `tempfile.TemporaryDirectory` with context manager to avoid it. Here is an example: https://docs.python.org/3/library/tempfile.html#examples

recite/models.py
govza: self is not defined. it works just like that.

PR #13, opened 2019-02-14
Reciter fixes

govza:
Fix some perfomance issues
Fix formatting and typos

### DIFF
--- a/recite/admin.py
+++ b/recite/admin.py
@@ -1,23 +1,33 @@
 import csv
 import io
 import json
+import os
+import shutil
+import tempfile
 from zipfile import ZipFile
 
 from django import forms
 from django.contrib import admin
+from django.core.files import File
+from django.forms import BaseInlineFormSet
 
 from quran.models import Ayah
 
 from .models import Recitation, Reciter
-import os
-import tempfile
-import shutil
-from django.core.files import File
+
+
+class RecitationFormSet(BaseInlineFormSet):
+    """FormSet for Recitation shows limited value of Recitations"""
+
+    def get_queryset(self):
+        qs = super(RecitationFormSet, self).get_queryset()
+        return qs[:10]
 
 
 class RecitationInline(admin.TabularInline):
     """Inline tab of Recitations in admin page of Reciter"""
     # Hide editing of single Recitation field
+
     def has_add_permission(self, request, obj=None):
         return False
 
@@ -26,8 +36,8 @@ class RecitationInline(admin.TabularInline):
 
     def has_delete_permission(self, request, obj=None):
         return False
-
     model = Recitation
+    formset = RecitationFormSet
 
 
 class ReciterForm(forms.ModelForm):
@@ -47,6 +57,7 @@ class ReciterAdmin(admin.ModelAdmin):
     Modified save_model method allows to save Recitations
     objects at the same model creation form
     """
+
     def has_change_permission(self, request, obj=None):
         return False
 
--- a/recite/admin.py
+++ b/recite/admin.py
@@ -142,10 +142,9 @@ class ReciterAdmin(admin.ModelAdmin):
 
         for row in csv_content_sorted:
             segments_list = json.loads(row.get("segments"))
-            segments = ",".join(
+            segments_string = ",".join(
                 (f"{segment[2]}:{segment[3]}" for segment in segments_list)
             )
-
             surah_str = row.get('sura')
             ayah_str = row.get('ayat')
             try:
@@ -153,7 +152,6 @@ class ReciterAdmin(admin.ModelAdmin):
                 ayah_number = int(ayah_str)
             except ValueError:
                 print(f"Not normal format for csv at row {row}")
-
-            segments_dict[surah_number][ayah_number] = segments_list
+            segments_dict[surah_number][ayah_number] = segments_string
 
         return segments_dict
--- a/recite/admin.py
+++ b/recite/admin.py
@@ -91,13 +91,13 @@ class ReciterAdmin(admin.ModelAdmin):
         shutil.rmtree(temp_dir)
 
     def get_file_paths(self, temp_dir):
-        """Return file paths from temp_dir checking subdirectories"""
-        file_paths = []
+        """Return file paths dict from temp_dir checking subdirectories"""
+        file_paths = {}
         for root, directories, files in os.walk(temp_dir):
             for filename in files:
                 if filename.endswith('.mp3'):
                     filepath = os.path.join(root, filename)
-                    file_paths.append(filepath)
+                    file_paths[filename] = filepath
 
         return file_paths
 
@@ -116,11 +116,7 @@ class ReciterAdmin(admin.ModelAdmin):
         """Search in uploaded files for corresponding ayah"""
 
         file_ayah_name_expected = str(ayah) + '.mp3'
-        file_ayah = None
-        for file_path in file_paths:
-            if file_ayah_name_expected == os.path.basename(file_path):
-                file_ayah = file_path
-                break
+        file_ayah = file_paths.get(file_ayah_name_expected)
 
         if not file_ayah:
             raise ValueError(
--- a/recite/admin.py
+++ b/recite/admin.py
@@ -47,6 +47,9 @@ class ReciterAdmin(admin.ModelAdmin):
     Modified save_model method allows to save Recitations
     objects at the same model creation form
     """
+    def has_change_permission(self, request, obj=None):
+        return False
+
     inlines = [
         RecitationInline,
     ]
--- a/recite/admin.py
+++ b/recite/admin.py
@@ -92,8 +92,9 @@ class ReciterAdmin(admin.ModelAdmin):
         file_paths = []
         for root, directories, files in os.walk(temp_dir):
             for filename in files:
-                filepath = os.path.join(root, filename)
-                file_paths.append(filepath)
+                if filename.endswith('.mp3'):
+                    filepath = os.path.join(root, filename)
+                    file_paths.append(filepath)
 
         return file_paths
 
--- a/recite/admin.py
+++ b/recite/admin.py
@@ -81,7 +81,7 @@ class ReciterAdmin(admin.ModelAdmin):
             audio_files.extractall(temp_dir)
 
             # let's create dict file_paths[surah][ayah] = file_path
-            file_paths = self.get_file_paths_dict(temp_dir=temp_dir)
+            file_paths = self.get_file_paths_dict(temp_dir)
             # let's create Recitations from segments and file_paths
             for surah_number, ayahs in segments_dict.items():
                 for ayah_number, segments in ayahs.items():
@@ -101,14 +101,14 @@ class ReciterAdmin(admin.ModelAdmin):
                             audio=file_ayah
                         )
 
-    def get_file_paths_dict(self, temp_dir):
+    def get_file_paths_dict(self, directory):
         """
         Process audio zip file uploaded and create dict
         :return: file_paths[surah][ayah] = file_path
         """
         file_paths = defaultdict(lambda: defaultdict(int))
 
-        for root, _directories, files in os.walk(temp_dir):
+        for root, _directories, files in os.walk(directory):
             for filename in files:
                 if filename.endswith('.mp3'):
                     filepath = os.path.join(root, filename)
--- a/recite/admin.py
+++ b/recite/admin.py
@@ -77,7 +77,7 @@ class ReciterAdmin(admin.ModelAdmin):
         # process audio zip file uploaded
         audio_zip_file = form.cleaned_data['audio_zip_file']
         temp_dir = self.get_unzipped_files_dir(audio_zip_file)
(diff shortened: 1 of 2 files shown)


recite/models.py
```diff
@@ -69,28 +68,9 @@ def __str__(self):
...
-    when corresponding `Reciter` object is  deleted.
+    when the corresponding `Reciter` object is deleted.
     """
     reciter_path = os.path.join(
         settings.MEDIA_ROOT, 'recite', str(instance.slug))
     if os.path.exists(reciter_path):
         shutil.rmtree(reciter_path)
-
-
-@receiver(models.signals.pre_save, sender=Reciter)
```
al-muammar, line 80: Why did you decide to delete this method?

recite/admin.py
```diff
@@ -88,12 +102,13 @@ def save_model(self, request, obj, form, change):
         shutil.rmtree(temp_dir)
 
     def get_file_paths(self, temp_dir):
```
al-muammar: Let's call it `get_audio_paths`?

recite/admin.py
```diff
@@ -88,12 +102,13 @@ def save_model(self, request, obj, form, change):
...
-        """Return file paths from temp_dir checking subdirectories"""
-        file_paths = []
+        """Return file paths dict from temp_dir checking subdirectories"""
+        file_paths = {}
```
al-muammar: Let's make it `audio_paths[surah][ayah]` dictionary, please.
You will hide the initial format here and all other parts won't need to know about file naming conventions.

recite/admin.py
```diff
@@ -112,11 +127,7 @@ def get_file_audio_path(self, file_paths, ayah):
...
         file_ayah_name_expected = str(ayah) + '.mp3'
-        file_ayah = None
-        for file_path in file_paths:
-            if file_ayah_name_expected == os.path.basename(file_path):
-                file_ayah = file_path
-                break
+        file_ayah = file_paths.get(file_ayah_name_expected)
 
         if not file_ayah:
             raise ValueError(
```
al-muammar:
```suggestion
            raise KeyError(
```

recite/admin.py
```diff
@@ -112,11 +127,7 @@ def get_file_audio_path(self, file_paths, ayah):
         """Search in uploaded files for corresponding ayah"""
 
         file_ayah_name_expected = str(ayah) + '.mp3'
-        file_ayah = None
-        for file_path in file_paths:
-            if file_ayah_name_expected == os.path.basename(file_path):
-                file_ayah = file_path
-                break
+        file_ayah = file_paths.get(file_ayah_name_expected)
```
al-muammar: All of this logic breaks once there is one missing file.
We need to find **all** missing files and output it to the user. Either do it in this PR, or create an issue to address it later.

al-muammar (changes requested):
A number of issues here:
1. You still didn't address the problem when exception happens after you extracted everything to tmp dir, but before you deleted it. It won't be cleaned. You have to use a context manager.
2. Have you profiled the problem of slowness? The only reason that I see now is copying files. It might be slow in case of 6k files. It's better to move them instead.

recite/models.py
govza: If we try to change Reciter in admin panel, the Recitations are recreated but the older ones are not deleted. It was like quick fix, really don't think these operations will be used often. Anyway, it possible to delete and recreate Reciter.

recite/admin.py
```diff
@@ -55,89 +69,91 @@ class ReciterAdmin(admin.ModelAdmin):
     def save_model(self, request, obj, form, change):
         super().save_model(request, obj, form, change)
 
-        # process csv file uploaded
-        segments_file = form.cleaned_data['segments_file'].file
-        csv_content = self.read_segments_csv(
-            segments_file, encoding=request.encoding)
+        # Create dict segments_dict[surah][ayah] = segments
```
al-muammar:
```suggestion
        # Create a dict segments_dict[surah][ayah] = segments
```

recite/admin.py
```diff
@@ -55,89 +69,91 @@ class ReciterAdmin(admin.ModelAdmin):
...
+        segments_dict = self.get_segments_dict(request=request, form=form)
 
-        # process audio zip file uploaded
+        # Extract audio files zip to temp directory
```
al-muammar:
```suggestion
        # Extract zipped audio files to a temporary directory
```

recite/admin.py
```diff
@@ -55,89 +69,91 @@ class ReciterAdmin(admin.ModelAdmin):
...
-
-                recitation, _ = Recitation.objects.get_or_create(
-                    ayah=ayah,
-                    reciter=obj,
-                    segments=segments,
-                    audio=file_ayah
-                )
-
-        shutil.rmtree(temp_dir)
+        audio_files = ZipFile(audio_zip_file, 'r')
```
al-muammar: Let's move it inside `with tempfile....`

recite/admin.py
```diff
@@ -55,89 +69,91 @@ class ReciterAdmin(admin.ModelAdmin):
...
+            ayah_str = row.get('ayat')
+            try:
+                surah_number = int(surah_str)
+                ayah_number = int(ayah_str)
+            except ValueError:
+                print(f"Not normal format for csv at row {row}")
 
-        temp_dir = tempfile.mkdtemp()
-        audio_files.extractall(temp_dir)
+            segments_dict[surah_number][ayah_number] = segments
```
al-muammar: Wait a second, why do you need a string representation? Why can't you use `segment_list`?

recite/admin.py
```diff
@@ -55,89 +69,91 @@ class ReciterAdmin(admin.ModelAdmin):
...
-
-        ayah = Ayah.objects.get(
-            number=ayah_id,
-            surah=surah_id,
-        )
-        return ayah
-
-    def get_file_audio_path(self, file_paths, ayah):
-        """Search in uploaded files for corresponding ayah"""
+    def get_segments_dict(self, request, form):
```
al-muammar: I would make this function more specific. Instead of giving `request` and `form` I would give only those parts of these objects that you are using here.

recite/admin.py
```diff
@@ -55,89 +69,91 @@ class ReciterAdmin(admin.ModelAdmin):
...
+                        )
+
+                        Recitation.objects.get_or_create(
+                            ayah=ayah,
+                            reciter=obj,
+                            segments=segments,
+                            audio=file_ayah
+                        )
+
+    def get_file_paths_dict(self, temp_dir):
```
al-muammar: Nit: it doesn't matter, is it a temporary directory or not. Call it just dir or directory

recite/admin.py
```diff
@@ -55,89 +69,91 @@ class ReciterAdmin(admin.ModelAdmin):
...
+
+        for root, _directories, files in os.walk(temp_dir):
             for filename in files:
-                filepath = os.path.join(root, filename)
-                file_paths.append(filepath)
+                if filename.endswith('.mp3'):
+                    filepath = os.path.join(root, filename)
+
+                    # extract surah and ayah number from filename
+                    surah_str = filename[:3]
```
al-muammar, line 134: Nit: I would extract it to a separate function and test it if necessary.

recite/admin.py
```diff
@@ -55,89 +69,91 @@ class ReciterAdmin(admin.ModelAdmin):
...
+                            segments=segments,
+                            audio=file_ayah
+                        )
+
+    def get_file_paths_dict(self, temp_dir):
+        """
+        Process audio zip file uploaded and create dict
+        :return: file_paths[surah][ayah] = file_path
+        """
+        file_paths = defaultdict(lambda: defaultdict(int))
```
al-muammar, line 126: Why do you need a defaultdict?

recite/admin.py
```diff
@@ -55,89 +69,91 @@ class ReciterAdmin(admin.ModelAdmin):
...
-        for root, directories, files in os.walk(temp_dir):
+        with tempfile.TemporaryDirectory() as temp_dir:
+            audio_files.extractall(temp_dir)
+
+            # let's create dict file_paths[surah][ayah] = file_path
+            file_paths = self.get_file_paths_dict(temp_dir=temp_dir)
+            # let's create Recitations from segments and file_paths
+            for surah_number, ayahs in segments_dict.items():
+                for ayah_number, segments in ayahs.items():
+                    with open(file_paths[surah_number][ayah_number], 'rb')\
```
al-muammar, line 90: What if there is no such file?

al-muammar (changes requested):
Now it's better, a number of small nits, however there are still few problems.
Secondly, can you add a section in the PR description about how you tested it.
Thanks!

recite/admin.py
govza: It should be standard zip file and it's single procedure of audio import done only by admins.

recite/admin.py
govza: Defaultdict returns None, no errors, no files added.

recite/admin.py
govza: We need surah_number, ayah_number and segments from csv file to create Recitation object. Seams elegant... and actually it was your proposal https://github.com/it-muslim/quran-words/pull/11#discussion_r256177494

recite/admin.py
al-muammar: I don't think so.
```py
>>> with open(None, 'rb') as fin:
...     pass
...
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

recite/admin.py
```diff
@@ -55,89 +69,91 @@ class ReciterAdmin(admin.ModelAdmin):
...
-        shutil.rmtree(temp_dir)
-
-    def get_file_paths(self, temp_dir):
-        """Return file paths from temp_dir checking subdirectories"""
-        file_paths = []
-        for root, directories, files in os.walk(temp_dir):
+        segments_dict = self.get_segments_dict(request.encoding, segments_file)
+
+        with tempfile.TemporaryDirectory() as temp_dir:
+            # Extract zipped audio files to temp directory
```
al-muammar: Nit
```suggestion
            # Extract zipped audio files to a temp directory
```

recite/admin.py
al-muammar: I mean that lines `117-123` should probably live in a separate function.

recite/admin.py
al-muammar: I'm saying that it should be as follows. You are breaking encapsulation in lines `145-147`.
```suggestion
            segments_dict[surah_number][ayah_number] = segments_list
```

recite/models.py
```diff
@@ -69,28 +68,9 @@ def __str__(self):
 def auto_delete_files_on_delete(sender, instance, **kwargs):
     """
     Deletes Recitation files from filesystem
```
al-muammar: Nit
```suggestion
    Deletes Recitation files from the filesystem
```

al-muammar (changes requested):
almost-there

recite/admin.py
govza: @Jihadik Strange, tested it manually with segments file with more lines than audio files in zip and vice versa.
Upd. Ok, was mistaken, defaultdict(int) returns 0, not None. Which makes no trouble.

recite/admin.py
govza: Commit breaks the import. We don't use list in Models, but it's converted to string representation (line 145).

recite/admin.py
govza: Returns None instead of KeyError if no files found, good way to initialize nested dict file_paths[surah][ayah]

recite/admin.py
```diff
@@ -55,89 +70,105 @@ class ReciterAdmin(admin.ModelAdmin):
...
+                (f"{segment[2]}:{segment[3]}" for segment in segments_list)
+            )
+            surah_str = row.get('sura')
+            ayah_str = row.get('ayat')
+            try:
+                surah_number = int(surah_str)
+                ayah_number = int(ayah_str)
+            except ValueError:
+                print(f"Not normal format for csv at row {row}")
+            segments_dict[surah_number][ayah_number] = segments_string
```
al-muammar, line 172: What will happen if you make the following change?
```suggestion
            segments_dict[surah_number][ayah_number] = segments_list
```

recite/admin.py
```diff
@@ -55,89 +70,105 @@ class ReciterAdmin(admin.ModelAdmin):
...
+                        )
+
+                        Recitation.objects.get_or_create(
+                            ayah=ayah,
+                            reciter=obj,
+                            segments=segments,
+                            audio=file_ayah
+                        )
+
+    def check_segments_has_files(self, segments_dict, file_paths):
```
al-muammar, line 106: Nice!

al-muammar (approved):
Approving the PR, however next time let's add a "test plan" section, so I can see how you tested your code.

recite/admin.py
govza: ValidationError at /admin/recite/reciter/add/

['List must contain tuples only']

PR #14, opened 2019-02-15
Quran api

govza:
Quran API implemented via Django-Rest-Framework
endpoints at 
http://127.0.0.1:8000/api/quran/surah/
http://127.0.0.1:8000/api/quran/surah/1

### DIFF
--- a/quran/tests.py
+++ b/quran/tests.py
@@ -31,7 +31,7 @@ class GetQuran(APITestCase):
     def test_get_ayahs_of_surahs(self):
         '''
         This test ensures that ayahs is loaded correctly to database
-        making a GET request to the quran/{surah_id} endpoint
+        making a GET request to the api/quran/surah/{number} endpoint
         '''
         # hit the API endpoint for each surah
         surahs = list(range(1, 115))
@@ -41,7 +41,7 @@ class GetQuran(APITestCase):
                 reverse(
                     "surah-details",
                     kwargs={
-                        "surah_id": surah_id
+                        "number": surah_id
                     }
                 )
             )
@@ -60,7 +60,7 @@ class GetQuran(APITestCase):
             reverse(
                 "surah-details",
                 kwargs={
-                    "surah_id": 115
+                    "number": 115
                 }
             )
         )
--- a/quran/tests.py
+++ b/quran/tests.py
@@ -8,49 +8,6 @@ from .serializers import SurahSerializer, AyahSerializer
 
 class GetQuran(APITestCase):
     '''Checking Quran app API'''
-    client = APIClient()
-
-    def test_get_surahs(self):
-        '''
-        This test ensures that surahs is loaded correctly to database
-        making a GET request to the quran/ endpoint
-        '''
-        # hit the API endpoint
-        response = self.client.get(
-            reverse("surahs-all")
-        )
-        # fetch the data from db
-        expected = Surah.objects.all()
-        serialized = SurahSerializer(expected, many=True)
-        self.assertEqual(response.data, serialized.data)
-        self.assertEqual(response.status_code, status.HTTP_200_OK)
-
-        # check number of Surahs
-        self.assertEqual(len(response.data), 114)
-
-    def test_get_ayahs_of_surahs(self):
-        '''
-        This test ensures that ayahs is loaded correctly to database
-        making a GET request to the api/quran/surah/{number} endpoint
-        '''
-        # hit the API endpoint for each surah
-        surahs = list(range(1, 115))
-
-        for surah_id in surahs:
-            response = self.client.get(
-                reverse(
-                    "surah-details",
-                    kwargs={
-                        "number": surah_id
-                    }
-                )
-            )
-            # fetch the data from db
-            expected = Ayah.objects.filter(surah=surah_id)
-            serialized = AyahSerializer(expected, many=True)
-
-            self.assertEqual(response.data['ayahs'], serialized.data)
-            self.assertEqual(response.status_code, status.HTTP_200_OK)
 
     def test_nonexistent_surah(self):
         '''
--- a/quran/tests.py
+++ b/quran/tests.py
@@ -1,0 +1,67 @@
+from django.urls import reverse
+from rest_framework.test import APIClient, APITestCase
+from rest_framework.views import status
+
+from .models import Surah, Ayah
+from .serializers import SurahSerializer, AyahSerializer
+
+
+class GetQuran(APITestCase):
+    '''Checking Quran app API'''
+    client = APIClient()
+
+    def test_get_surahs(self):
+        '''
+        This test ensures that surahs is loaded correctly to database
+        making a GET request to the quran/ endpoint
+        '''
+        # hit the API endpoint
+        response = self.client.get(
+            reverse("surahs-all")
+        )
+        # fetch the data from db
+        expected = Surah.objects.all()
+        serialized = SurahSerializer(expected, many=True)
+        self.assertEqual(response.data, serialized.data)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        # check number of Surahs
+        self.assertEqual(len(response.data), 114)
+
+    def test_get_ayahs_of_surahs(self):
+        '''
+        This test ensures that ayahs is loaded correctly to database
+        making a GET request to the quran/{surah_id} endpoint
+        '''
+        # hit the API endpoint for each surah
+        surahs = list(range(1, 115))
+
+        for surah_id in surahs:
+            response = self.client.get(
+                reverse(
+                    "surah-details",
+                    kwargs={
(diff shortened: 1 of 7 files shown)


quran/tests.py
```diff
@@ -0,0 +1,67 @@
...
+
+from .models import Surah, Ayah
+from .serializers import SurahSerializer, AyahSerializer
+
+
+class GetQuran(APITestCase):
+    '''Checking Quran app API'''
+    client = APIClient()
+
+    def test_get_surahs(self):
```
al-muammar: Tbh, I don't feel that we need this test. You are just testing `ListAPIView` correctness, which should be tested in DRF, not here.
You are also testing here a loading to DB. It should be tested in somewhere in models or in fixtures.

quran/views.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class SurahDetailsView(generics.RetrieveAPIView):
+    """
+    Provides a get method handler for Surah details.
+    """
+    queryset = Surah.objects.all()
+    serializer_class = serializers.SurahSerializer
+
+    def get(self, request, *args, **kwargs):
```
al-muammar: You don't need to do it. Redefine just `get_queryset` method. It should be enough.
You can also use `get_object_or_404` method from `django.shortcuts`.

quran/urls.py
```diff
@@ -0,0 +1,17 @@
+'''Quran app api url endpoints'''
+from django.urls import re_path
+
+from . import views
+
+urlpatterns = [
+    re_path(
+        r'quran/(?P<surah_id>\d+)/$',
```
al-muammar: Why do you use regular expressions? I think it's an old style ;)

quran/tests.py
```diff
@@ -0,0 +1,67 @@
...
+            reverse("surahs-all")
+        )
+        # fetch the data from db
+        expected = Surah.objects.all()
+        serialized = SurahSerializer(expected, many=True)
+        self.assertEqual(response.data, serialized.data)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        # check number of Surahs
+        self.assertEqual(len(response.data), 114)
```
al-muammar: Let's define `114` as a constant beforehand.

quran/tests.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def test_get_ayahs_of_surahs(self):
```
al-muammar: The same as `test_get_surahs`

quran/tests.py
```diff
@@ -0,0 +1,67 @@
...
+                )
+            )
+            # fetch the data from db
+            expected = Ayah.objects.filter(surah=surah_id)
+            serialized = AyahSerializer(expected, many=True)
+
+            self.assertEqual(response.data['ayahs'], serialized.data)
+            self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+    def test_nonexistent_surah(self):
```
al-muammar, line 12: I believe, this is a good test. It actually tests some not DB specific logic. Let's keep it here.
However, will it work before I initialize a DB?

al-muammar (changes requested):
You are on a good track, keep going!

quran/tests.py
```diff
@@ -0,0 +1,67 @@
...
+        '''
+        This test ensures that ayahs is loaded correctly to database
+        making a GET request to the quran/{surah_id} endpoint
+        '''
+        # hit the API endpoint for each surah
+        surahs = list(range(1, 115))
+
+        for surah_id in surahs:
+            response = self.client.get(
+                reverse(
```
An4ik: some  suggestions:

use `def get_url(surah_number)`   and set URL for this API endpoint i.g.:
`
def get_url(surah_number):
    return reverse(...)
`


Also please always try to use [typing hints](https://docs.python.org/3/library/typing.html) since we use Python3 
 

So your code may look like this:

`
def get_url(surah_number: int) -> str:
    return reverse(...)
`


 For example here: 
  `surahs = list(range(1, 115))`

I've expected that it's list of the names.

quran/tests.py
```diff
@@ -0,0 +1,67 @@
...
+                    "surah-details",
+                    kwargs={
+                        "surah_id": surah_id
+                    }
+                )
+            )
+            # fetch the data from db
+            expected = Ayah.objects.filter(surah=surah_id)
+            serialized = AyahSerializer(expected, many=True)
```
An4ik: the name should be `serializer`

quran/tests.py
```diff
@@ -0,0 +1,67 @@
...
+            response = self.client.get(
+                reverse(
+                    "surah-details",
+                    kwargs={
+                        "surah_id": surah_id
+                    }
+                )
+            )
+            # fetch the data from db
+            expected = Ayah.objects.filter(surah=surah_id)
```
An4ik: the name should be `qs`  or `queryset`

quran/tests.py
```diff
@@ -0,0 +1,67 @@
...
+            serialized = AyahSerializer(expected, many=True)
+
+            self.assertEqual(response.data['ayahs'], serialized.data)
```
An4ik: serialized.data return dict, so 
you may user self.assertCountEqual (cause we use python3)

quran/urls.py
```diff
@@ -0,0 +1,17 @@
...
+        views.SurahDetailsView.as_view(),
+        name="surah-details"
+    ),
+    re_path(
+        r'quran/$',
+        views.SurahListView.as_view(),
+        name="surahs-all",
```
An4ik: instead of surahs-all it shoul be surahs-list

quran/views.py
```diff
@@ -0,0 +1,36 @@
+from rest_framework import generics
+from rest_framework.response import Response
+from rest_framework.views import status
+
+from . import serializers
+from .models import Surah
+
+
```
An4ik, line 6: use this `class SurahView(GenericAPIView, RetrieveModelMixin, ListModelMixin):`. So instead of two views you may have only one.

look at the ListApiView   it inherits from GenericAPIView and ListModelMixin
look at then SurahDetailsView  it inherits from  GenericAPIView and [RetrieveModelMixin](https://www.django-rest-framework.org/api-guide/generic-views/#retrievemodelmixin)

so

quran/urls.py
An4ik: `quran`  that's GREAT book, but bad name for our situation. Please rename it or add /surah/
so it's more readable when we have `quran/surah/` or just `surah`

quran/views.py
An4ik: I guess you don't need this method at all.
Once you've inherited from `generics.RetrieveAPIView` that's enough. It's done in backgoround.
just add to class attribute  ` lookup_field = 'surah_id'` (by default it's looking for `pk` or `id` so if you change
 r'quran/(?P<surah_id>\d+)/$',   surah_id to pk then no need to define lookup_field.

Just google about it (lookup field DRF etc) you wil find more info.

quran/tests.py
An4ik: Sura -  it's single sura.
Surah - means more than one sura. so no need to add `s` to the tail.
https://translate.google.com/?source=gtx_c#view=home&op=translate&sl=auto&tl=en&text=%D1%81%D1%83%D1%80%D0%B0%0A%D1%81%D1%83%D1%80%D1%8B

govza: @Jihadik I'm not quite sure about issue here. For me it's too early to separate test_models.py test_drf.py etc. This is test of DRF, if I understood it right, basically [following this tutorial.](https://medium.com/backticks-tildes/lets-build-an-api-with-django-rest-framework-32fcf40231e5)

govza: @An4ik Thank you, good question. Asked my friends arabophones, the explanation is not that easy. 
Surah is one of the irregular words when it becomes plural. Although it is female, and we always add “t” to females word to make it plural , the plural of Surah is sowar. 

However, here we try to [use english](https://en.wikipedia.org/wiki/S%C5%ABrah), if it's possible in naming of variables.

al-muammar: @govza, in a couple of words: these tests are useless, because you are testing libraries, not your logic.

quran/tests.py
al-muammar: These are good suggestions, however, since we don't have types at all, let's introduce type hints in a separate PR.

quran/tests.py
al-muammar: @govza, I'm simply saying, just delete these tests. You are trying to test some library functions which is unnecessary.

quran/tests.py
al-muammar: In general I agree, however, since we don't have type hints in the whole project, let's do it in a separate PR and then require type annotations in all further PRs.

al-muammar (approved):
Delete unnecessary tests and you are good to go.
Accepting the PR in order to unblock the progress.
spada_accept_to_unblock

PR #15, opened 2019-03-02
Improvement/refactored q api close #8 issue

### LINKED ISSUE #8: Rewrite all views in class-based fasion
The modern way of writing views is based on classes, so we need to use this style everywhere.

An4ik:
updated from Quran api branch  and did some refactoring of the code.  It has the same functionality as before.


quran/api/views.py
```diff
@@ -0,0 +1,17 @@
...
+    """View that allows to List, Retrieve Surah."""
+    lookup_field = 'number'
+    queryset = Surah.objects.all()
+    serializer_class = SurahListSerializer
+
+    def get_serializer_class(self):
+        """Returns appreciate serializer class for the given action."""
+        if self.action == "retrieve":
+            return SurahDetailsSerializer
+        return super().get_serializer_class()
```
An4ik, line 22: Actually this returns `serializer_class` which we set on view (SurahListSerializer). Yeap we also can write

`return SurahListSerializer` instead of   `return super().get_serializer_class()`. But by this way we show that it's overwritten function.

config/urls.py
```diff
@@ -1,8 +1,14 @@
-"""Project URL Configuration"""
+from django.conf.urls import url
 from django.contrib import admin
+from rest_framework import routers
 from django.urls import path, include
 
+from quran.api import views as quran_views
+
+router = routers.SimpleRouter()
```
al-muammar, line 8: What's the benefit of it?

quran/api/tests/serializers/test_ayah_serializer.py
```diff
@@ -0,0 +1,12 @@
...
+from quran.api.serializers import AyahSerializer
+
+
+class TestAyahSerializer(SimpleTestCase):
+    def test_fields(self):
+        expected_fileds = (
+            'number',
+            'text',
+        )
+        self.assertCountEqual(AyahSerializer().get_fields().keys(), expected_fileds)
```
al-muammar: What's the value in this test? Aren't you just testing DRF, which is unnecessary?
Additionally, once you decided to serialize other fields, this test will fail, which will require you to change it. That's not ideal.

quran/api/tests/serializers/test_surah_details_serializer.py
```diff
@@ -0,0 +1,14 @@
...
+
+class TestSurahDetailsSerializer(SimpleTestCase):
+    def test_fields(self):
+        expected_fileds = (
+            "number",
+            "name",
+            "total_ayahs",
+            "ayahs",
+        )
+        self.assertCountEqual(SurahDetailsSerializer().get_fields().keys(), expected_fileds)
```
al-muammar: The same goes here, you are testing DRF which is unnecessary.

quran/api/tests/serializers/test_surah_list_serializer.py
```diff
@@ -0,0 +1,13 @@
+from django.test import SimpleTestCase
+
+from quran.api.serializers import SurahListSerializer
+
+
+class TestSurahListSerializer(SimpleTestCase):
+    def test_fields(self):
```
al-muammar: Let's remove it too?

quran/api/tests/views/test_surah_list_retrieve_view.py
```diff
@@ -0,0 +1,35 @@
...
+
+class TestSurahListRetrieveView(APITestCase):
+
+    def setUp(self):
+        self.list_url = reverse("api:quran-list")
+        self.surah_number = 1
+        self.details_url = reverse("api:quran-detail", kwargs={"number": self.surah_number})
+
+    def test_list_url(self):
+        self.assertEqual(self.list_url, "/api/quran/surah/")
```
al-muammar: What's the values of the test?
You are using reverse urls in order to avoid hard-coding them. And here you are hard-coding them. So once the url changes, your tests are broken.
On the other hand, I can see the logic of protecting yourself from changing the API by mistake.

It's quite controversial.

quran/api/tests/views/test_surah_list_retrieve_view.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def test_details_url(self):
+        self.assertEqual(self.details_url, "/api/quran/surah/1/")
+
+    def test_list(self):
+        response = self.client.get(self.list_url)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(len(response.data), 114)
```
al-muammar: Make 114 a special constant.

quran/api/views.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class SurahListRetrieveView(viewsets.ReadOnlyModelViewSet):
+    """View that allows to List, Retrieve Surah."""
+    lookup_field = 'number'
+    queryset = Surah.objects.all()
+    serializer_class = SurahListSerializer
+
+    def get_serializer_class(self):
+        """Returns appreciate serializer class for the given action."""
```
al-muammar: What serializer?

quran/api/views.py
```diff
@@ -0,0 +1,17 @@
+from rest_framework import viewsets
+
+from ..models import Surah
+from .serializers import SurahListSerializer, SurahDetailsSerializer
+
+
+class SurahListRetrieveView(viewsets.ReadOnlyModelViewSet):
+    """View that allows to List, Retrieve Surah."""
```
al-muammar: Nit
```suggestion
    """View that allows to list and retrieve Surah."""
```

quran/api/views.py
al-muammar, line 22: This is a bit confusing. `SurahListSerializer` and `SurahDetailsSerializer` are not entirely symmetric, which makes me think that there is a special logic behind that.
Let's rewrite it without specifying `serializer_class`.

Regarding your comment, I'm not entirely sure, how much value showing override gives. You just have to know that there are special functions. You might indicate it in a comment.

al-muammar (changes requested):
palpatine_good

quran/api/serializers.py
```diff
@@ -1,27 +1,27 @@
...
         fields = ('number', 'text')
 
 
-class SurahDetailsSerializer(serializers.ModelSerializer):
-    '''
-    Serializer for Surah details which
-    include Surah information and its Ayahs
-    '''
+class SuraDetailsSerializer(serializers.ModelSerializer):
+    """Serializer for detailed Sura detailed."""
```
govza, line 23: Typo?

config/urls.py
An4ik: > REST framework adds support for automatic URL routing to Django, and provides with a simple, quick and consistent way of wiring your view logic to a set of URLs.

config/urls.py
```diff
@@ -1,8 +1,14 @@
...
+router.register(r'quran/surah', quran_views.SurahListRetrieveView, base_name="quran")
+
 urlpatterns = [
     path('admin/', admin.site.urls),
-    path('', include('quran.urls')),
+    url(r'^api/', include((router.urls, 'api'), namespace='api')),
```
al-muammar: I don't think we need regexes here: https://docs.djangoproject.com/en/2.1/topics/http/urls/#using-regular-expressions

al-muammar (approved):
whatever

config/urls.py
An4ik: oooh, yeap.

govza:
Is it common way to make nested directories with /api/views etc ?

govza:
@An4ik ready to merge? You are not in discord group?

An4ik:
@govza  yes, it's ready.

PR #16, opened 2019-03-10
Recitation media

govza:
Media available by link on server, fixed bug with wrong file paths.
... Special thanks goes to
###  [Test plan](https://github.com/it-muslim/quran-words/wiki/Tests)
 - Basic django functions
 - Adding Reciter

### DIFF
--- a/config/urls.py
+++ b/config/urls.py
@@ -2,6 +2,7 @@ from django.conf.urls import url
 from django.contrib import admin
 from rest_framework import routers
 from django.urls import path, include
+from django.conf import settings
 
 from quran.api import views as quran_views
 
@@ -12,3 +13,8 @@ urlpatterns = [
     path('admin/', admin.site.urls),
     url('api/', include((router.urls, 'api'), namespace='api')),
 ]
+
+if settings.DEBUG:
+    from django.conf.urls.static import static
+    urlpatterns += static(
+        settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
--- a/config/urls.py
+++ b/config/urls.py
@@ -1,13 +1,15 @@
-from django.conf.urls import url
-from django.contrib import admin
-from rest_framework import routers
-from django.urls import path, include
 from django.conf import settings
+from django.conf.urls import url
+from django.conf.urls.static import static
+from django.contrib import admin
+from django.urls import include, path
+from rest_framework import routers
 
 from quran.api import views as quran_views
 
 router = routers.SimpleRouter()
-router.register(r'quran/surah', quran_views.SurahListRetrieveView, base_name="quran")
+router.register(
+    r'quran/surah', quran_views.SurahListRetrieveView, base_name="quran")
 
 urlpatterns = [
     path('admin/', admin.site.urls),
@@ -15,6 +17,5 @@ urlpatterns = [
 ]
 
 if settings.DEBUG:
-    from django.conf.urls.static import static
     urlpatterns += static(
         settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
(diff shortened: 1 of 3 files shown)


config/urls.py
```diff
@@ -1,8 +1,14 @@
 """Project URL Configuration"""
 from django.contrib import admin
 from django.urls import path, include
+from django.conf import settings
```
al-muammar, line 1: Nice, I didn't know that you could export setting in such a nice way :)

recite/admin.py
```diff
@@ -167,8 +167,8 @@ def get_segments_dict(self, encoding, segments_file):
             try:
                 surah_number = int(surah_str)
                 ayah_number = int(ayah_str)
+                segments_dict[surah_number][ayah_number] = segments_string
```
al-muammar: I like the previous way more, because you show that the exception may happen only in two lines, rather than three.
It was more clear.

config/urls.py
```diff
@@ -1,8 +1,14 @@
...
 
 urlpatterns = [
     path('admin/', admin.site.urls),
     path('', include('quran.urls')),
 ]
+
+if settings.DEBUG:
+    from django.conf.urls.static import static
```
al-muammar: Tbh, I don't like this kind of imports :)
I prefer moving everything to the top for visibility.

al-muammar (commented):
Can it be covered with unit test by any mean?
goodcatch

recite/admin.py
govza: Actually the bug was here.
try: file is bismillah.mp3  -> exception occured but it was written to file_paths with surah_number and ayah_number of previous loop.
media file was named 002001 with real file bismillah.mp3
Problem is we are not stopping import because of theese file, just printing them to output

config/urls.py
govza: Don't think this import will have any influence on prod, so no problem.

config/urls.py
govza: [Django docs](https://docs.djangoproject.com/en/2.1/howto/static-files/#serving-files-uploaded-by-a-user-during-development) brought me here.

govza:
Could be covered by unit tests within docker, by creating superuser and importing fixture files, not easy though.

recite/admin.py
govza: Ok, think the proper way is to use try, except, **else**

recite/admin.py
```diff
@@ -169,6 +169,6 @@ def get_segments_dict(self, encoding, segments_file):
                 ayah_number = int(ayah_str)
             except ValueError:
                 print(f"Not normal format for csv at row {row}")
-            segments_dict[surah_number][ayah_number] = segments_string
-
+            else:
```
al-muammar, line 183: I rarely use `else` in exceptions, but it looks fine.
I just understood what was the problem.

Since we found a bug, let's add a unit test (or couple of them).

al-muammar (changes requested):
canihazunittest

al-muammar (changes requested):
Good, however, where is the unit test? :)
canihazunittest

recite/tests/test_admin.py
```diff
@@ -0,0 +1,51 @@
...
+        Test that function returns only valid surah ayah strucured dict.
+
+        Parse mocked directory and return structured dict of filenames,
+        `file_paths[surah_number][ayah_number] = filepath`
+        Ignore malformed filenames.
+        """
+        with mock.patch('os.walk') as mock_walk:
+            # mocked os.walk returns fake directory
+            mock_walk.return_value = [
+                ('tmp', ['Al-Test'], []),
```
al-muammar, line 30: :)

al-muammar (approved):
Well done!

PR #18, opened 2019-03-30
Production deployment uses nginx and gunicorn to serve static data.

govza:
Production deployment uses nginx and gunicorn to serve static data.

Added nginx conatiner,
gunicorn dependencies,
Updated launch command to auto collect static, make migrations and run gunicorn.
Nginx config, prevents timeout within 10min for /admin to manage recitations

[Tested](https://github.com/it-muslim/quran-words/wiki/Tests) with full reciter packages addition.

### DIFF
--- a/docker-compose-win.yml
+++ b/docker-compose-win.yml
@@ -7,11 +7,7 @@ services:
     container_name: web01
     env_file:
       - ./.env
-    command: >
-            sh -c "python3 manage.py collectstatic --no-input &&
-            python3 manage.py makemigrations &&
-            python3 manage.py migrate &&
-            gunicorn config.wsgi -b 0.0.0.0:8000 --workers 3 --timeout 600"
+    entrypoint: "/www/entrypoint.sh"
     volumes:
       - .:/www
     ports:
--- a/docker-compose-win.yml
+++ b/docker-compose-win.yml
@@ -4,7 +4,7 @@ version: "3"
 services:
   web:
     build: .
-    container_name: web01
+    container_name: web
     env_file:
       - ./.env
     entrypoint: "/www/entrypoint.sh"
@@ -16,7 +16,7 @@ services:
       - db
 
   db:
-    container_name: db01
+    container_name: db
     image: "postgres:10.5-alpine"
     restart: always
     env_file:
@@ -27,7 +27,7 @@ services:
       - postgres:/var/lib/postgresql/data
 
   adminer:
-    container_name: adminer01
+    container_name: adminer
     image: adminer
     restart: always
     ports:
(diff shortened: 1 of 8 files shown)


docker-compose-win.yml
```diff
@@ -0,0 +1,56 @@
+# before run use command `docker volume create postgres`
+# run docker-compose -f docker-compose-win.yml up
+version: "3"
+services:
+  web:
+    build: .
+    container_name: web01
+    env_file:
+      - ./.env
+    command: >
```
al-muammar: Create an `entrypoint.sh` and put it there.

docker-compose-win.yml
```diff
@@ -0,0 +1,56 @@
+# before run use command `docker volume create postgres`
+# run docker-compose -f docker-compose-win.yml up
+version: "3"
+services:
+  web:
+    build: .
+    container_name: web01
```
al-muammar: Why do you add `01` suffix to everything?

docker-compose-win.yml
```diff
@@ -0,0 +1,56 @@
...
+    env_file:
+      - ./.env
+    command: >
+            sh -c "python3 manage.py collectstatic --no-input &&
+            python3 manage.py makemigrations &&
+            python3 manage.py migrate &&
+            gunicorn config.wsgi -b 0.0.0.0:8000 --workers 3 --timeout 600"
+    volumes:
+      - .:/www
+    ports:
```
al-muammar: Remove it, you don't need to open this port. At most you need to expose it.
The same everywhere except nginx.

al-muammar (changes requested):
Why do you need two `docker-compose` files?
Otherwise, looks pretty decent.

govza:
docker-compose-win used to postgresql [bug workaround](https://github.com/docker/for-win/issues/445) for windows, it creates external volume. 
For linux it's not needed,  postgresql data is linked to volume /postgres/

docker-compose-win.yml
govza: It's one of our futures swarm :)

entrypoint.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+python3 manage.py collectstatic --no-input
+python3 manage.py makemigrations
+python3 manage.py migrate
+gunicorn config.wsgi -b 0.0.0.0:8000 --workers 3 --timeout 600
```
al-muammar: Try the following thing and then specify your command in `docker-compose`. In this way you don't need two entry points for development and production.
```suggestion
exec "$@"
```

al-muammar (approved):

Imangazaliev (approved):

